PR #7, opened 2017-08-25
Update poll stopping logic in load

aheyman11:
If the Poller is looking at the integer representing sleep, only sleep if there are additional relations sitting in the `to_poll` queue.

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -678,9 +678,9 @@ def create_source_tables_when_ready(relations: List[LoadableRelation], max_concu
                     raise ETLRuntimeError(
                         "No new extracts found in last %s seconds, bailing out" % idle_termination_seconds)
                 else:
-                    logger.debug("Poller: Sleeping for %s seconds", item)
-                    time.sleep(item)
                     if to_poll.qsize():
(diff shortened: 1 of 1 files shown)


python/etl/load.py
```diff
@@ -678,9 +678,9 @@ def poll_worker():
                     raise ETLRuntimeError(
                         "No new extracts found in last %s seconds, bailing out" % idle_termination_seconds)
                 else:
-                    logger.debug("Poller: Sleeping for %s seconds", item)
-                    time.sleep(item)
                     if to_poll.qsize():
```
blohr-hs: I'm probably being dense, but I don't quite follow the logic here. Shouldn't this test be true if there *are* relations waiting in the queue, and therefore the sleep will only occur if the queue is approximately non-empty? From the commit comment, it seemed like you wanted it to sleep if the queue was empty.

aheyman11: You're totally right -- my commit message was backwards. But I think the logic of what we actually want is correct.

blohr-hs: OK, cool, thank you!

PR #8, opened 2017-08-30
DW-198 Retry support for extract

blohr-hs:
This add support for extract to retry on a per-table basis. The number of retry attempts it uses is controlled by a new configuration setting, and has a default value of 1 retry.

This was tested out using the `FakeSqoopExtractor`; so it does work in the case where Sqoop returns a non-zero exit code. 

Open questions/issues for feedback:
* Are there any known failure cases today where Sqoop returns a non-zero error code and we do *not* want to retry? The current implementation is a very blunt instrument, and just always retries Sqoop.
* The Static and ManifestOnly extractors both simply interact with S3 via Boto. Boto internally handles retries for various transient errors, so it didn't seem necessary to add explicit retries in user code on top of that. Thoughts?
* The Spark extractor seems to be broken, based on the comments in `spark.py` on lines 156 and 157. Because of that, the complexity involved in creating a test environment I could use to test retry code for Spark errors, and not finding some readily available documentation on the exceptions/errorrs thrown by Spark, I didn't test this out on Spark. Should Spark be out of scope for this?

### DIFF
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -79,6 +79,8 @@ def _build_config_map(settings):
     for section in {"object_store", "resources", "etl_events"}.intersection(settings):
         for name, value in _flatten_hier(section, settings[section]):
             mapping[name] = value
+    for top_level_scalar_key in {"extract_retries"}.intersection(settings):
+        mapping[top_level_scalar_key] = settings[top_level_scalar_key]
     return mapping
 
 
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -76,11 +76,9 @@ def _flatten_hier(prefix, props):
 
 def _build_config_map(settings):
     mapping = OrderedDict()
-    for section in {"object_store", "resources", "etl_events"}.intersection(settings):
+    for section in {"arthur_settings", "object_store", "resources", "etl_events"}.intersection(settings):
         for name, value in _flatten_hier(section, settings[section]):
             mapping[name] = value
-    for top_level_scalar_key in {"extract_retries"}.intersection(settings):
-        mapping[top_level_scalar_key] = settings[top_level_scalar_key]
     return mapping
 
 
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -1,4 +1,9 @@
 {
+    # General settings controlling how Arthur itself behaves
+    "arthur_settings": {
+        # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries
+        "extract_retries": 1
+    },
     # Target (Redshift) cluster
     "data_warehouse": {
         # The environment variable must contain a full connection string for an admin user to create a database.
@@ -71,7 +76,5 @@
             # The bytea data type is probably not useful, but we'll try to pull it in base64 format.
             "bytea": ["varchar(65535)", "encode(%s, 'base64')", "string"]
         }
-    },
-    # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times.
-    "extract_retries": 1
+    }
 }
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -72,6 +72,6 @@
             "bytea": ["varchar(65535)", "encode(%s, 'base64')", "string"]
         }
     },
-    # If an extract from an upstream source fails for any reason, retry the extract at most this many times.
+    # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times.
     "extract_retries": 1
 }
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -71,5 +71,7 @@
             # The bytea data type is probably not useful, but we'll try to pull it in base64 format.
             "bytea": ["varchar(65535)", "encode(%s, 'base64')", "string"]
         }
-    }
+    },
+    # If an extract from an upstream source fails for any reason, retry the extract at most this many times.
+    "extract_retries": 1
 }
--- a/python/etl/config/settings.schema
+++ b/python/etl/config/settings.schema
@@ -98,6 +98,17 @@
     },
     "type": "object",
     "properties": {
+        "arthur_settings": {
+            "type": "object",
+            "properties": {
+                "extract_retries": {
+                    "description": "If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries",
+                    "type": "integer",
+                    "minimum": 0
+                }
+            },
+            "additionalProperties": false
+        },
         "data_warehouse": {
             "type": "object",
             "properties": {
@@ -345,11 +356,6 @@
         "required_in_full_load": {
             "description": "DEPRECATED USE 'required_for_success' instead",
             "$ref": "#/definitions/glob_pattern_list"
-        },
-        "extract_retries": {
-            "description": "If an extract from an upstream source fails due to some transient error, retry the extract at most this many times.",
-            "type": "integer",
-            "minimum": 1
         }
     },
     "required": [ "data_warehouse", "sources", "type_maps" ],
--- a/python/etl/config/settings.schema
+++ b/python/etl/config/settings.schema
@@ -347,6 +347,7 @@
             "$ref": "#/definitions/glob_pattern_list"
         },
         "extract_retries": {
+            "description": "If an extract from an upstream source fails due to some transient error, retry the extract at most this many times.",
             "type": "integer",
             "minimum": 1
         }
--- a/python/etl/config/settings.schema
+++ b/python/etl/config/settings.schema
@@ -107,6 +107,7 @@
                     "minimum": 0
                 }
             },
+            "required": [ "extract_retries" ],
             "additionalProperties": false
         },
         "data_warehouse": {
@@ -358,6 +359,6 @@
             "$ref": "#/definitions/glob_pattern_list"
         }
     },
-    "required": [ "data_warehouse", "sources", "type_maps" ],
+    "required": [ "arthur_settings", "data_warehouse", "sources", "type_maps" ],
     "additionalProperties": false
 }
--- a/python/etl/config/settings.schema
+++ b/python/etl/config/settings.schema
@@ -345,6 +345,10 @@
         "required_in_full_load": {
             "description": "DEPRECATED USE 'required_for_success' instead",
             "$ref": "#/definitions/glob_pattern_list"
+        },
+        "extract_retries": {
+            "type": "integer",
+            "minimum": 1
         }
     },
     "required": [ "data_warehouse", "sources", "type_maps" ],
--- a/python/etl/errors.py
+++ b/python/etl/errors.py
@@ -4,6 +4,19 @@ class ETLError(Exception):
     """
 
 
+class PermanentETLError(ETLError):
+    """
+    Represents all types of errors that are known to be permanent failures due irrecoverable situations within the ETL
+    """
+
+
+class TransientETLError(ETLError):
+    """
+    Represents all types of errors that could potentially resolve themselves due to changes in the outside environment
+    or some kind of intervention, and are thus retryable.
+    """
+
+
 class ETLSystemError(ETLError):
     """
     Blunder in the ETL code -- you'll need to check the ETL code, sorry about that.
@@ -136,6 +149,11 @@ class SqoopExecutionError(DataExtractError):
     pass
 
 
+class TransientSqoopExecutionError(TransientETLError):
+    # TODO: Refactor exception hierarchy to properly tag all transient errors
+    pass
+
+
 class MissingCsvFilesError(DataExtractError):
     pass
 
@@ -195,4 +213,13 @@ class RequiredRelationLoadError(ETLRuntimeError):
 class DataUnloadError(ETLRuntimeError):
     """
     Exception when the unload operation fails
+    """
+
+
+class RetriesExhaustedError(PermanentETLError):
+    """
+    Raised when all retry attempts have been exhausted.
+
+    The causing exception should be passed to this one to complete the chain of failure accountability. For example:
+    >>> raise RetriesExhaustedError from ETLRuntimeError("Boom!")
     """
--- a/python/etl/errors.py
+++ b/python/etl/errors.py
@@ -225,7 +225,7 @@ class RetriesExhaustedError(PermanentETLError):
     """
 
 
-def retry(max_attempts: int, callback: callable, logger=None):
+def retry(max_retries: int, callback: callable, logger):
     """
     Retry a function a maximum number of times and return its results.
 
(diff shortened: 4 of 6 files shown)


python/etl/errors.py
```diff
@@ -196,3 +214,43 @@ class DataUnloadError(ETLRuntimeError):
...
+    """
+    Retry a function a maximum number of times and return its results.
+
+    The given callback function is only retried if it throws a TransientETLError. Any other error is considered
+    permanent, and therefore no retry attempt is made.
+    """
+    failure_reason = None
+    successful_result = None
+
+    for attempt in range(max_attempts + 1):
```
blohr-hs: rename to max_retreis

mikekaminsky: i'd prefer `max_retries` ;)

python/etl/config/__init__.py
```diff
@@ -79,6 +79,8 @@ def _build_config_map(settings):
     for section in {"object_store", "resources", "etl_events"}.intersection(settings):
         for name, value in _flatten_hier(section, settings[section]):
             mapping[name] = value
+    for top_level_scalar_key in {"extract_retries"}.intersection(settings):
```
tvogels01: Instead of adding scalars at the top level (where we'll have to touch this line of code for every new one), let's just have a new section like `resources` or `etl_events`.  If nothing better comes to mind, `general_settings` or `arthur_settings` works.  It'll be the section for settings of algorithms inside Arthur like number of retries for load and extract or the number of seconds before timeout in extract.

python/etl/config/settings.schema
```diff
@@ -345,6 +345,11 @@
         "required_in_full_load": {
             "description": "DEPRECATED USE 'required_for_success' instead",
             "$ref": "#/definitions/glob_pattern_list"
+        },
+        "extract_retries": {
+            "description": "If an extract from an upstream source fails due to some transient error, retry the extract at most this many times.",
+            "type": "integer",
+            "minimum": 1
```
tvogels01: I'd prefer a `minimum` of `0` as it should be possible to turn retries off, no matter how useful we think this is.  The number of retries becomes, to some degree, the feature flag.

python/etl/errors.py
```diff
@@ -4,6 +4,19 @@ class ETLError(Exception):
     """
 
 
+class PermanentETLError(ETLError):
+    """
+    Represents all types of errors that are known to be permanent failures due irrecoverable situations within the ETL
+    """
+
+
+class TransientETLError(ETLError):
```
tvogels01: Please move this under `ETLRuntimeError` since the other high-level errors, like configuration or code errors, won't be re-tryable.  Only runtime errors should be inspected and retried.

python/etl/errors.py
```diff
@@ -4,6 +4,19 @@ class ETLError(Exception):
...
+    """
+    Represents all types of errors that could potentially resolve themselves due to changes in the outside environment
+    or some kind of intervention, and are thus retryable.
```
tvogels01: The "some kind of intervention" contradicts our idea of retrying automatically.  The scope for the decision of transient vs. permanent should be the same ETL run.

python/etl/errors.py
```diff
@@ -136,6 +149,11 @@ class SqoopExecutionError(DataExtractError):
     pass
 
 
+class TransientSqoopExecutionError(TransientETLError):
```
tvogels01: Is the existing SqoopExecutionError no longer used?  Why keep it?

Also, why step outside the `DataExtractError`? An entirely separate, parallel hierarchy just so that we can tag transient errors seems trouble on the one side.  On the other, using multiple inheritance might keep the hierarchy flatter. An error while executing Sqoop should be both, DataExtractError and transient ... so `class SqoopExecutionError(DataExtractError, TransientETLError)`?

Also, please add a comment to new exception classes when you get a chance instead of the `pass`.

python/etl/errors.py
```diff
@@ -196,3 +214,43 @@ class DataUnloadError(ETLRuntimeError):
...
     Exception when the unload operation fails
     """
+
+
+class RetriesExhaustedError(PermanentETLError):
+    """
+    Raised when all retry attempts have been exhausted.
+
+    The causing exception should be passed to this one to complete the chain of failure accountability. For example:
+    >>> raise RetriesExhaustedError from ETLRuntimeError("Boom!")
```
tvogels01: Please use `run_tests.py` to make sure your doc tests work.  This one doesn't ... it's missing the output from the exception.  See for example https://github.com/harrystech/arthur-redshift-etl/blob/f0ffbe98e5186d0a1c675de884886970af107c4a/python/etl/names.py#L353

python/etl/errors.py
```diff
@@ -196,3 +214,43 @@ class DataUnloadError(ETLRuntimeError):
...
+    """
+
+
+def retry(max_retries: int, callback: callable, logger):
```
tvogels01: I wouldn't look for the retry code in the `errors.py` module.  This file was created to capture (only) the error hierarchy.

python/etl/errors.py
```diff
@@ -196,3 +214,43 @@ class DataUnloadError(ETLRuntimeError):
...
+    The given callback function is only retried if it throws a TransientETLError. Any other error is considered
+    permanent, and therefore no retry attempt is made.
+    """
+    failure_reason = None
+    successful_result = None
+
+    for attempt in range(max_retries + 1):
+        try:
+            successful_result = callback(attempt)
+        except TransientETLError as e:
```
tvogels01: Is there a reason to not just use `as failure_reason`?

python/etl/errors.py
```diff
@@ -196,3 +214,43 @@ class DataUnloadError(ETLRuntimeError):
...
+    """
+    Retry a function a maximum number of times and return its results.
+
+    The given callback function is only retried if it throws a TransientETLError. Any other error is considered
+    permanent, and therefore no retry attempt is made.
+    """
+    failure_reason = None
+    successful_result = None
+
+    for attempt in range(max_retries + 1):
```
tvogels01: Very nit-picky here ... it's really `max_tries` since the first try is not a re-try. :-)

python/etl/extract/extractor.py
```diff
@@ -12,8 +12,14 @@
...
 import etl.s3
 import etl.db
+from etl.config import get_config_value
 from etl.config.dw import DataWarehouseSchema
-from etl.errors import DataExtractError, ETLRuntimeError, MissingCsvFilesError
+from etl.errors import (
+    DataExtractError,
+    ETLRuntimeError,
+    MissingCsvFilesError,
+    retry
```
tvogels01: This is where I noticed that `retry` is a bit out of place inside `errors`.

python/etl/extract/extractor.py
```diff
@@ -71,15 +77,21 @@ def extract_source(self, source: DataWarehouseSchema,
...
+                                                              'object_key': relation.manifest_file_name},
+                                                 index={"current": i + 1, "final":
+                                                        len(relations), "name": source.name},
+                                                 dry_run=self.dry_run,
+                                                 attempt_num=attempt_num + 1):
+                                self.extract_table(source, relation)
+
+                    retry(get_config_value("extract_retries"), _monitored_table_extract, self.logger)
+
                 except ETLRuntimeError:
```
tvogels01: Since the current transient (and permanent) errors stepped outside of the existing `ETLRuntimeError` hierarchy, the failures in extraction would no longer be properly accounted for.

python/etl/config/settings.schema
```diff
@@ -98,6 +98,17 @@
...
+        "arthur_settings": {
+            "type": "object",
+            "properties": {
+                "extract_retries": {
+                    "description": "If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries",
+                    "type": "integer",
+                    "minimum": 0
+                }
+            },
+            "additionalProperties": false
```
blohr-hs: Should this section and the `extract_retries` key be marked as required? From the end user's perspective, they're not required due to their presence in `default_settings.py`, but it is needed by the code, so I wasn't sure which way to go there.

python/etl/errors.py
blohr-hs: The `TransientSqoopExecutionError` wasn't intended to live beyond this PR. I just needed to add it to make the code work until we can hash out what the exception hierarchy should look like. The multiple inheritance route seems the most appealing at this point; I'll look into that.

python/etl/config/settings.schema
tvogels01: Yes.  Because they are in the default settings, we can make them required :-)  Having reasonable defaults makes the code much easier since you now don't have to test for the presence of the value.  (There is some old code lingering around that deals with optional values only for backwards compatibility.)

python/etl/extract/extractor.py
```diff
@@ -71,15 +77,22 @@ def extract_source(self, source: DataWarehouseSchema,
...
+                                                 source=self.source_info(source, relation),
+                                                 destination={'bucket_name': relation.bucket_name,
+                                                              'object_key': relation.manifest_file_name},
+                                                 index={"current": i + 1, "final":
+                                                        len(relations), "name": source.name},
+                                                 dry_run=self.dry_run,
+                                                 attempt_num=attempt_num + 1):
+                                self.extract_table(source, relation)
+
+                    retries = int(get_config_value("arthur_settings.extract_retries"))  # type: ignore
```
blohr-hs: Mypy seemed very unhappy with the int conversion here. Given that the config schema specifies this to be an int, it seems reasonably safe. Also, by local mypy environment seems to be broken; it throws the following error every time, which isn't in our user code:

`venv/lib/mypy/typeshed/stdlib/3/calendar.pyi:27: error: Name 'List' is not defined`

python/etl/errors.py
blohr-hs: Based on that explanation, isn't `max_retries` correct then, because the first try is already excluded (that's the "+ 1" part)?

python/etl/errors.py
blohr-hs: I originally had it that way, but changed it to the current way to make it more explicit and not have the local variable overwritten as a side effect of catching the exception, which seemed too magical and non-obvious.

python/etl/extract/extractor.py
blohr-hs: Yeah, I didn't see another module that seemed like a better home, so I stuck it there. Open to suggestions on where it should live.

python/etl/errors.py
```diff
@@ -122,7 +134,7 @@ class UpstreamValidationError(ETLRuntimeError):
     """
 
 
-class DataExtractError(ETLRuntimeError):
+class DataExtractError(TransientETLError):
```
blohr-hs: Currently, there isn't anything that catches `DataExtractError`s, so it seems pretty safe, and it's in-line with the general intention to retry extract errors.

python/etl/config/default_settings.yaml
```diff
@@ -1,4 +1,9 @@
 {
+    # General settings controlling how Arthur itself behaves
+    "arthur_settings": {
+        # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries
+        "extract_retries": 1
```
mikekaminsky: Should we have a "delay" parameter here that allows one to specify how long to wait between retries?

blohr-hs: It would certainly be easy enough to add if you guys thought it was necessary. Most of the extract jobs are relatively slow to start anyway because they either spin up big jobs or do a lot of network IO, so there's some inherent delay. Thoughts?

python/etl/extract/extractor.py
tvogels01: Yes, happens to me too. I can't find a reference as to when this started breaking in typeshed.

python/etl/config/default_settings.yaml
mikekaminsky: @tvogels01 has a better sense for these things then I do, and I defer to him. I was thinking about the "exponential back-off" that I'd implemented elsewhere and realized we hadn't had any discussion along those lines for this feature.

tvogels01: Leaning towards not parameterizing it for now.  It would be nice to get to a state where we can attribute these fixes (or additional parameters) to specific incidents ... if it never breaks, why build in the contingency?
We do a poor job with testing today (my bad), but another way to look at this is that for every parameter that you add, you need to write tests and documentation. Which should cut down on extraneous code :-)

mikekaminsky: "failure driven development"

python/etl/errors.py
tvogels01: 👍

python/etl/extract/extractor.py
tvogels01: For now, let's leave it like this. No reason to get fancy unless we have further use ... I could see us eventually having a `get_int_config_value`. Out of scope.

python/etl/extract/extractor.py
tvogels01: Let's leave it here for now.  Should we decide to grow the functionality, we can refactor then.

python/etl/errors.py
```diff
@@ -196,3 +208,46 @@ class DataUnloadError(ETLRuntimeError):
...
+
+    The causing exception should be passed to this one to complete the chain of failure accountability. For example:
+    >>> raise RetriesExhaustedError from ETLRuntimeError("Boom!")
+    Traceback (most recent call last):
+        ...
+    etl.errors.RetriesExhaustedError
+    """
+
+
+def retry(max_retries: int, callback: Callable[[int], T], logger) -> T:
```
tvogels01: Chris will be so happy with all the type annotations :-)  Nifty.

python/etl/errors.py
tvogels01: yep, got confused between attempts, tries, etc.

python/etl/errors.py
```diff
@@ -196,3 +208,46 @@ class DataUnloadError(ETLRuntimeError):
...
+                logger.warning("Encountered the following error (retrying %s more times): %s",
+                               max_retries - attempt, str(e))
+            continue
+        except:
+            # We consider all other errors permanent and immediately re-raise without retrying
+            raise
+        else:
+            break
+    else:
+        raise RetriesExhaustedError from failure_reason
```
tvogels01: Final suggestion ... adding number of retries here makes it easier to see what the parameter was. E.g.
```
raise RetriesExhaustedError("reached max number of retries ({:d})".format(max_retries)) from failure_reason
```

(Also and maybe, the general preference is to raise an instance, not a class?)

tvogels01:
RE Spark -- yes, it's out of scope since Spark extracts are only used for debugging. The spark extract isn't really that broken ... there are some edge cases where the load would fail.  Meaning that using Spark extract to test any logic of this step locally is still valuable.  (At least, to me.)

python/etl/config/default_settings.yaml
tvogels01: "failure driven exception handling"  You can easily get yourself into a spot where you add all sorts of error handling.  But that code is never exercised in tests or production and then becomes actually the brittle part of the system.

PR #11, opened 2017-09-08
Change logging text for continue_from

bhtucker:
Was confused using this, as "matching" in the Arthur relation context normally means 'glob matching', but here meant 'exact match'


tvogels01:
Thanks.  This is a hold-over from where I thought mistakenly that a glob would be a good idea.

PR #13, opened 2017-09-11
DW-227 Extract from older version of PostgreSQL (in particular Redshift)

bhtucker:
Relatively shallow integration of Redshift as an upstream data source.

User-facing changes:
* Arthur now supports any data source that can be extracted using the postgresql JDBC driver, in particular Redshift
* For non-postgres sources, tables are pessimistically assumed to be size 671 mb for the purposes of efficiently partitioning during extract.

### DIFF
--- a/python/etl/extract/spark.py
+++ b/python/etl/extract/spark.py
@@ -79,7 +79,7 @@ class SparkExtractor(DatabaseExtractor):
         partition_key = relation.find_partition_key()
 
         with closing(etl.db.connection(source.dsn, readonly=True)) as conn:
-            table_size = self.fetch_source_table_size(conn, relation)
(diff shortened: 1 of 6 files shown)


python/etl/extract/spark.py
```diff
@@ -78,9 +78,10 @@ def read_table_as_dataframe(self, source: DataWarehouseSchema, relation: Relatio
         """
         partition_key = relation.find_partition_key()
 
+        table_size = self.fetch_source_table_size(source.dsn, relation)
+        num_partitions = self.maximize_partitions(table_size)
+
         with closing(etl.db.connection(source.dsn, readonly=True)) as conn:
```
tvogels01: This `with` statement should now be right above the `self.determine_partitioning`.  No point in opening the connection now if the `predicates` are left at `None`.

PR #15, opened 2017-09-14
Render and Combine Templates

bhtucker:
Putting this up for discussion —

Provide pipeline definition combination command for more modular pipeline definitions

TODO:
* type annotations
* refactor pipeline definitions and installers
* potentially clarify argparse groups/arguments definitions for render_template

### DIFF
--- a/python/etl/render_template/splice.py
+++ b/python/etl/render_template/splice.py
@@ -1,0 +1,29 @@
+import logging
+
+import yaml
+
+import etl.config
+import etl.render_template
+
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+
+def splice_pipeline(base_pipeline_template, new_template, insertion_point='PingCronutAfterBootstrap', replace=False,
+                    compact=False):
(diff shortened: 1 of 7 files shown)


tvogels01:
I like the idea of assembling our pipeline definitions.  This would make it much easier to keep them consistent.  I'm not sure that I'm seeing this here.  Can we draw this up on a whiteboard?

python/etl/render_template/splice.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+
+def splice_pipeline(base_pipeline_template, new_template, insertion_point='PingCronutAfterBootstrap', replace=False,
+                    compact=False):
+    base_pipeline = yaml.safe_load(etl.render_template._render_template_string(base_pipeline_template)[0])
+    splice_pipeline = yaml.safe_load(etl.render_template._render_template_string(new_template)[0])
```
blohr-hs: This would be a lot clearer with tuple unpacking or something instead of a random index magic number.

tvogels01:
Moving discussion to Issue #21

PR #16, opened 2017-09-14
Support logging for mark_failure outside of except blocks

bhtucker:
Currently, we may get a logging error about lacking exception context


tvogels01:
Looks good.  In a more perfect world, we could save the exception context ... not now.

PR #25, opened 2017-09-21
Hotfix - add mypy to requirements to avoid failing import of typing

tvogels01:
This happened after playing with too many Python versions.  Our production uses 3.4 which doesn't have `typing`.  Next version of ETL moves to Python 3.5 which does have `typing`


tvogels01:
This ran successfully in development.

PR #30, opened 2017-10-02
Optionally template S3 data sources

bhtucker:
allow s3 source schema definitions to reference arthur config


tvogels01:
@bhtucker This needs a corresponding change in the settings schema which currently expects an identifier for the bucket name and doesn't accept a template string.

PR #35, opened 2017-10-08
Add terminate_backends command

tvogels01:
Sometimes we encounter sessions that have transactions with locks that interfere with the ETL.
This command allows to terminate those.

@bhtucker @triplec1988 Should we make this part of the rebuild ETL?


bhtucker:
Pushed up a more aggressive query that I think we need for 'un-sticking' promote

PR #38, opened 2017-10-11
Support / in prefix for SNS topics

aheyman11:
Allow for `/` in prefix names when installing pipelines.

We add an attribute `sns_prefix` to the object store, which is the same as `s3.prefix` but with `/` replaced with `_`. Then we make sure to reference that attribute in all pipeline templates.

Do the same thing when creating SNS topics in `sns_subscribe.sh`.

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -116,6 +116,7 @@ def run_arg_as_command(my_name="arthur.py"):
             setattr(args, "bucket_name", etl.config.get_config_value("object_store.s3.bucket_name"))
             if hasattr(args, "prefix"):
                 etl.config.set_config_value("object_store.s3.prefix", args.prefix)
+                etl.config.set_config_value("object_store.sns_prefix", args.prefix.replace('/', '_'))
                 if getattr(args, "use_monitor"):
                     etl.monitor.set_environment(args.prefix)
 
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -116,7 +116,7 @@ def run_arg_as_command(my_name="arthur.py"):
             setattr(args, "bucket_name", etl.config.get_config_value("object_store.s3.bucket_name"))
             if hasattr(args, "prefix"):
                 etl.config.set_config_value("object_store.s3.prefix", args.prefix)
-                etl.config.set_config_value("object_store.sns_prefix", args.prefix.replace('/', '_'))
+                etl.config.set_config_value("resources.SNS.sns_suffix", args.prefix.replace('/', '_'))
                 if getattr(args, "use_monitor"):
                     etl.monitor.set_environment(args.prefix)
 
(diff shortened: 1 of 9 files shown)


python/etl/commands.py
```diff
@@ -116,6 +116,7 @@ def run_arg_as_command(my_name="arthur.py"):
             setattr(args, "bucket_name", etl.config.get_config_value("object_store.s3.bucket_name"))
             if hasattr(args, "prefix"):
                 etl.config.set_config_value("object_store.s3.prefix", args.prefix)
+                etl.config.set_config_value("object_store.sns_prefix", args.prefix.replace('/', '_'))
```
tvogels01: Why should the SNS topic be part of the object store?

python/etl/templates/pizza_load_pipeline.json
```diff
@@ -22,7 +22,7 @@
         },
         {
             "id": "SNSParent",
-            "topicArn": "arn:aws:sns:${resources.VPC.region}:${resources.VPC.account}:redshift-etl-status_${object_store.s3.prefix}"
+            "topicArn": "arn:aws:sns:${resources.VPC.region}:${resources.VPC.account}:redshift-etl-status_${object_store.sns_prefix}"
```
tvogels01: Shouldn't the topic for page get the same treatment?

aheyman11: Yes indeed.

python/etl/commands.py
aheyman11: No strong reason – would `resources.SNS.prefix` be better?

bhtucker:
Seems reasonable under a name like `sns.suffix`.

This does diverge from how we've avoided this problem in `/current` and `/validation` cases, and how we sanitize table names for DynamoDB. I'd be more comfortable with this if we have a clear philosophy around nested prefixes in general.

To some extent this problem is emerging because of sharing one `object_store` across multiple Arthur projects, so is this problem a signal that we really shouldn't be doing that?

tvogels01:
Indeed, the DynamoDB naming is what came to my mind.  We should sanitize the same way across AWS resources.

As for the object store, we could split this up for our two lines of work.  We should probably keep the same data lake and so make sure we don't step on each others toes there.  But then data lakes should be combined so that seems fair.

In general, I'm not thrilled about consistency of prefixes (or lack thereof). Maybe we do need to add the project in, say `dev` or `prod` or `tom` or `andrea`?  So the SNS topic would be `dw-etl-dev-status_development` where `dw-etl-dev` is shared across multiple resources to make it easy to search?  (So it would be the `dw-etl-dev-events_development` table in DynamoDB then.)

tvogels01:
Discussions aside, I'd be happy to merge into v1.5 RC for now and we can go from there.

aheyman11:
Up to you @tvogels01 on whether you want to merge this... I had to hack something in order to install pipelines using our shared object store, so I mainly put this PR up in case it was useful to others. Definitely open to waiting until we conform this with Dynamo.

tvogels01:
I'm leaning towards making two changes.  One is to use `-` as the separator. Or maybe even re-use the code from DynamoDB which is a lot more paranoid.  The other change is to call this the "topic suffix" since SNS has more than topics so "sns suffix" is not descriptive enough.

PR #40, opened 2017-10-11
Run terminate backends in dry-run during rebuild

tvogels01:
Let's start using terminate_backends during the rebuild and make it go live if it seems ok.

### DIFF
--- a/python/etl/data_warehouse.py
+++ b/python/etl/data_warehouse.py
@@ -288,52 +288,44 @@ def create_new_user(config, new_user, group=None, add_user_schema=False, skip_us
                 etl.db.alter_search_path(conn, user.name, search_path)
 
 
-def list_transactions(cx) -> List[int]:
+def list_open_transactions(cx):
     """
-    Print information about sessions and tables that they have a lock on.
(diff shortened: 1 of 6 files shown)


python/etl/data_warehouse.py
```diff
@@ -288,49 +288,64 @@ def create_new_user(config, new_user, group=None, add_user_schema=False, skip_us
                 etl.db.alter_search_path(conn, user.name, search_path)
 
 
-def list_transactions(cx):
+def list_transactions(cx) -> List[int]:
+    """
+    Print information about sessions and tables that they have a lock on.
+    Return PIDs of the backends for each session.
+
+    This skips sessions of the user that is owning the connection.
```
blohr-hs: It might be worth explicitly documenting the fact that this function returns a list of the transaction PIDs in addition to printing the report (despite the return type being shown as List). This function is basically both a query for the list of open transaction PIDs, as well as a command to print a report about the open transactions.

tvogels01: This reads like a suggestion to split into `list_transactions` and `print_transactions` functions (which is what I now did)

PR #57, opened 2017-11-03
Fix: Update tag to look for when listing pipelines

aheyman11:
To conform with newer pipeline definitions.

Look for a tag with key `user:project` and value `data-warehouse` instead of a `DataWarehouseEnvironment` tag.


tvogels01:
That's fixed here: https://github.com/harrystech/arthur-redshift-etl/pull/56/files#diff-653885d55e0957fc45643e5051b2557c in PR #56

aheyman11:
Oops. Should have finished reading that PR first!

PR #60, opened 2017-11-21
Ignore unmanaged schemas

aheyman11:
Defines the concept of "managed" schemas to be those specified in your DW config, and allow Arthur to ignore unmanaged schemas 1) when determining dependency ordering, and 2) when loading a relation with unmanaged dependencies in staging mode.

The most prominent change is that RelationDescription dependencies are now TableName objects instead of strings, and those objects have an `is_managed` property that references `etl.config`.

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -112,7 +112,10 @@ class LoadableRelation:
         the position (staging or not) of a table.
 
         >>> import etl.file_sets
-        >>> fs = etl.file_sets.TableFileSet(TableName("a", "b"), TableName("c", "b"), None)
+        >>> from etl.config.dw import DataWarehouseSchema
+        >>> target_table_name = TableName("c", "b")
+        >>> target_table_name._base_schemas = [DataWarehouseSchema(schema_info={'name': 'c', 'owner': ''})]
+        >>> fs = etl.file_sets.TableFileSet(TableName("a", "b"), target_table_name, None)
         >>> relation = LoadableRelation(RelationDescription(fs), {}, skip_copy=True)
         >>> "As delimited identifier: {:s}, as string: {:x}".format(relation, relation)
         'As delimited identifier: "c"."b", as string: \\'c.b\\''
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -188,7 +188,7 @@ class LoadableRelation:
         if self.use_staging:
             # Rewrite the query to use staging schemas:
             for dependency in self.dependencies:
-                staging_dependency = TableName.from_identifier(dependency).as_staging_table_name()
+                staging_dependency = dependency.as_staging_table_name()
                 stmt = re.sub(r'\b' + dependency + r'\b', staging_dependency.identifier, stmt)
         return stmt
 
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -1077,7 +1077,7 @@ def show_upstream_dependencies(relations: List[RelationDescription], selector: T
     """
     List the relations upstream (towards sources) from the selected ones, report in execution order.
     """
-    execution_order = etl.relation.order_by_dependencies(relations)
+    execution_order = etl.relation.order_by_dependencies(relations, selector.base_schemas)
     selected_relations = etl.relation.find_matches(execution_order, selector)
     if len(selected_relations) == 0:
         logger.warning("Found no matching relations for: %s", selector)
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -1077,7 +1077,7 @@ def show_upstream_dependencies(relations: List[RelationDescription], selector: T
     """
     List the relations upstream (towards sources) from the selected ones, report in execution order.
     """
-    execution_order = etl.relation.order_by_dependencies(relations, selector.base_schemas)
+    execution_order = etl.relation.order_by_dependencies(relations)
     selected_relations = etl.relation.find_matches(execution_order, selector)
     if len(selected_relations) == 0:
         logger.warning("Found no matching relations for: %s", selector)
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -112,10 +112,12 @@ class LoadableRelation:
         the position (staging or not) of a table.
 
         >>> import etl.file_sets
-        >>> from etl.config.dw import DataWarehouseSchema
-        >>> target_table_name = TableName("c", "b")
-        >>> target_table_name._base_schemas = [DataWarehouseSchema(schema_info={'name': 'c', 'owner': ''})]
-        >>> fs = etl.file_sets.TableFileSet(TableName("a", "b"), target_table_name, None)
+        >>> import etl.config
+        >>> from collections import namedtuple
+        >>> MockDWConfig = namedtuple('MockDWConfig', ['schemas'])
+        >>> MockSchema = namedtuple('MockSchema', ['name'])
+        >>> etl.config._dw_config = MockDWConfig(schemas=[MockSchema(name='c')])
+        >>> fs = etl.file_sets.TableFileSet(TableName("a", "b"), TableName("c", "b"), None)
         >>> relation = LoadableRelation(RelationDescription(fs), {}, skip_copy=True)
         >>> "As delimited identifier: {:s}, as string: {:x}".format(relation, relation)
         'As delimited identifier: "c"."b", as string: \\'c.b\\''
--- a/python/etl/names.py
+++ b/python/etl/names.py
@@ -68,6 +68,7 @@ class TableName:
 
     Comparisons (for schema and table names) are case-insensitive.
 
+    >>> from etl.config.dw import DataWarehouseSchema
     >>> orders = TableName.from_identifier("www.orders")
     >>> str(orders)
     '"www"."orders"'
@@ -86,6 +87,7 @@ class TableName:
     >>> purchases = TableName.from_identifier("www.purchases")
     >>> orders < purchases
     True
+    >>> purchases._base_schemas = [DataWarehouseSchema(schema_info={'name': 'www', 'owner': ''})]
     >>> staging_purchases = purchases.as_staging_table_name()
     >>> staging_purchases.table == purchases.table
     True
@@ -93,17 +95,18 @@ class TableName:
     False
     """
 
-    __slots__ = ("_schema", "_table", "_staging")
+    __slots__ = ("_schema", "_table", "_staging", "_base_schemas")
 
     def __init__(self, schema: Optional[str], table: str) -> None:
         # Concession to subclasses ... schema is optional
         self._schema = schema.lower() if schema else None
         self._table = table.lower()
         self._staging = False
+        self._base_schemas = [] # type: List
 
     @property
     def schema(self):
-        # for system table dependencies, the schema should not be in a "staging" version
+        # for unmanaged dependencies, the schema should not be in a "staging" version
         if self.staging and self.is_managed:
             return as_staging_name(self._schema)
         else:
@@ -141,8 +144,9 @@ class TableName:
 
     @property
     def is_managed(self) -> bool:
-        base_schemas = etl.config.get_dw_config().schemas
-        return self._schema in [s.name for s in base_schemas]
+        if not self._base_schemas:
+            self._base_schemas = etl.config.get_dw_config().schemas
+        return self._schema in [s.name for s in self._base_schemas]
 
     @classmethod
     def from_identifier(cls, identifier: str):
@@ -161,7 +165,9 @@ class TableName:
         """
         Delimited table identifier to safeguard against unscrupulous users who use "default" as table name...
 
+        >>> from etl.config.dw import DataWarehouseSchema
         >>> tn = TableName("hello", "world")
+        >>> tn._base_schemas = [DataWarehouseSchema(schema_info={'name': 'hello', 'owner': ''})]
         >>> str(tn)
         '"hello"."world"'
         >>> str(tn.as_staging_table_name())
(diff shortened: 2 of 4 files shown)


python/etl/names.py
```diff
@@ -84,25 +87,28 @@ class TableName:
...
     >>> staging_purchases = purchases.as_staging_table_name()
     >>> staging_purchases.table == purchases.table
     True
     >>> staging_purchases.schema == purchases.schema
     False
     """
 
-    __slots__ = ("_schema", "_table", "_staging")
+    __slots__ = ("_schema", "_table", "_staging", "_base_schemas")
+    _identifier_slots = ("_schema", "_table", "_staging")
```
bhtucker: It seems like we use `.identifier` as a sufficient hash key — can we drop `_staging` out of here and then use these slots in `to_tuple`?

python/etl/names.py
```diff
@@ -84,25 +87,28 @@ class TableName:
...
 
     def __init__(self, schema: Optional[str], table: str) -> None:
         # Concession to subclasses ... schema is optional
         self._schema = schema.lower() if schema else None
         self._table = table.lower()
         self._staging = False
+        self._base_schemas = None  # type: Optional[List]
```
bhtucker: If we only care to do string containment checks, we should go for a set of strings

python/etl/names.py
```diff
@@ -137,6 +149,10 @@ def identifier(self) -> str:
         """
         return "{}.{}".format(*self.to_tuple())
 
+    @property
+    def is_managed(self) -> bool:
+        return self._schema in [s.name for s in self.base_schemas]
```
bhtucker: In keeping with the set-of-strings idea, let `base_schemas` access the `name` property

python/etl/names.py
```diff
@@ -242,6 +260,7 @@ def match_pattern(self, pattern: str) -> bool:
     def as_staging_table_name(self):
         tn = TableName(*self.to_tuple())
         tn._staging = True
+        tn._base_schemas = self._base_schemas
```
bhtucker: We do save a little execution time by copying this, but I'd advocate for the simplicity of only setting the private property mentioned in the method name

python/etl/relation.py
```diff
@@ -188,9 +188,10 @@ def query_stmt(self) -> str:
         return str(self._query_stmt)  # The str(...) shuts up the type checker.
 
     @property
-    def dependencies(self) -> FrozenSet[str]:
+    def dependencies(self) -> FrozenSet[TableName]:
         if self._dependencies is None:
-            self._dependencies = frozenset(self.table_design.get("depends_on", []))
+            dependent_table_names = [TableName.from_identifier(d) for d in self.table_design.get("depends_on", [])]
+            self._dependencies = frozenset(dependent_table_names)
```
bhtucker: This saves us a `TableName` instantiation in the SQL rewrite for staging

python/etl/relation.py
```diff
@@ -401,7 +404,7 @@ def order_by_dependencies(relation_descriptions):
         minimum, tie_breaker, description = queue.get()
         if minimum > 2 * nr_tables:
             raise CyclicDependencyError("Cannot determine order, suspect cycle in DAG of dependencies")
-        others = [table_map[dep].order for dep in description.dependencies]
+        others = [table_map[dep.identifier].order for dep in description.dependencies if dep.is_managed]
```
bhtucker: We can keep the pattern of sets/dicts of `TableName`s if we make `table_map` out of `target_table_name`s

python/etl/load.py
```diff
@@ -112,7 +112,10 @@ def __format__(self, code):
         the position (staging or not) of a table.
 
         >>> import etl.file_sets
-        >>> fs = etl.file_sets.TableFileSet(TableName("a", "b"), TableName("c", "b"), None)
+        >>> from etl.config.dw import DataWarehouseSchema
+        >>> target_table_name = TableName("c", "b")
+        >>> target_table_name._base_schemas = [DataWarehouseSchema(schema_info={'name': 'c', 'owner': ''})]
```
tvogels01: Once we start using `_base_schemas` outside the `names` module, accessing something private is not so great ...
At a minimum, we could add a setter:
```
    @base_schemas.setter
    def base_schemas(self, schemas: List) -> None:
        self._base_schemas = schemas
```
Theoretically, we could make sure that the base_schemas get set only once (which is the assumption in code, I believe).  But practically, this feels outside the scope.

python/etl/names.py
tvogels01: On a separate yet related note ... I'm not a fan of `_identifier_slots`.  The `__slots__` tuple is more of a setting than a field as it is a hint to Python which field names will be used. But the "identifier slots" are really more of a class constant (and are used only once). So it's more `HASH_FIELDS` or `UNIQUE_IDENTIFIER_FIELDS`?

Could we just use `to_tuple`? (`hash(self.to_tuple())`?) Conceptually, could we ever talk about tables that are both in staging and not at the same time? If the answer is yes, then simply using `__slots__` in the `hash` method is semantically meaningful. (This last point is how I read Benson's comment.)

python/etl/names.py
tvogels01: Not quite.  Please keep this as a list. The "base schemas" also allow sorting and make the list of all tables predictable (which is nicely hidden in the selector code.) Using a set would kill that since changing the config would alter the order of schemas when printed. If we're worried about performance, we could use an `OrderedDict`.

Overall (my) expectation is that the list of relations is ordered by the schemas in the way they are ordered in the configuration and then alphabetically by table name. Since sources come before transformations, this keeps us from mingling source and transformation schemas in ordering.

python/etl/names.py
```diff
@@ -84,25 +87,28 @@ class TableName:
...
         self._schema = schema.lower() if schema else None
         self._table = table.lower()
         self._staging = False
+        self._base_schemas = None  # type: Optional[List]
 
     @property
     def schema(self):
-        # for system table dependencies, the schema should not be in a "staging" version
-        if self.staging and not self._schema.startswith('pg_catalog'):
+        # for unmanaged dependencies, the schema should not be in a "staging" version
```
tvogels01: This comment should probably start going into the documentation of the class. This interface of "staging" versions is official enough now (and getting more complex) to warrant an explanation in the class doc.

python/etl/names.py
```diff
@@ -115,6 +121,12 @@ def table(self):
     def staging(self):
         return self._staging
 
+    @property
+    def base_schemas(self) -> List:
+        if self._base_schemas is None:
+            self._base_schemas = etl.config.get_dw_config().schemas
+        return self._base_schemas or []  # the 'or []' is just for the type checker
```
tvogels01: Torn between "nifty" and "horrified" :-)   Actually, this expresses our idea that "empty means anything" so wouldn't be there just for the type checker. Going with "nifty".

python/etl/names.py
tvogels01: Curious ... if "base schemas" are not set then everything is "not managed"? This changes past assumptions. Ok, but worth pointing out.

(We'll need something like "any schemas not defined as source or transformation in the configuration will be assumed to be outside of Arthur's control and thus not managed.")

python/etl/names.py
tvogels01: This deservers a comment (doc string) to understand that his is a cloning method.

Purists might argue that we should have something like a `staging: Optional[bool]=False` argument in the constructor to make it easier to see what's going on. Then this method would simply be (along with Benson's observation):
```
def as_staging_table_name(self):
   return TableName(*self.to_tuple(), staging=True)
```

python/etl/relation.py
```diff
@@ -364,7 +365,7 @@ def order_by_dependencies(relation_descriptions):
     RelationDescription.load_in_parallel(relation_descriptions)
     descriptions = [SortableRelationDescription(description) for description in relation_descriptions]
 
-    known_tables = frozenset({description.identifier for description in descriptions})
+    known_tables = frozenset({d.target_table_name for d in relation_descriptions})
```
tvogels01: Is there a reason to go from `description` to `d`? I find code that has the meaning spelled out for me so much more readable. And bits are mostly free these days ;-)
Either
```
    known_tables = frozenset({description.target_table_name for description in relation_descriptions})
```
or
```
    known_tables = frozenset({relation.target_table_name for relation in relation_descriptions})
```
is harder to write but much easier to read in the long run. This is true here and in a couple other places in this PR (for some value of "true" in my own opinion).

python/etl/relation.py
```diff
@@ -373,25 +374,27 @@ def order_by_dependencies(relation_descriptions):
     known_unknowns = set()
     queue = PriorityQueue()
     for initial_order, description in enumerate(descriptions):
-        pg_internal_dependencies = set(d for d in description.dependencies if d.startswith('pg_catalog'))
-        unknowns = description.dependencies - known_tables - pg_internal_dependencies
+        # superset including internal dependencies
+        unmanaged_dependencies = set(d for d in description.dependencies if not d.is_managed)
+        pg_internal_dependencies = set(d for d in description.dependencies if d.schema == 'pg_catalog')
```
tvogels01: If the lack of symmetry offends you, an `is_in_catalog` property would help. (I'm fine with as-is.)

python/etl/relation.py
tvogels01: Along that line ... should `target_table_name` be a field in a sortable relation description? Reaching into the "original description" feels a bit wrong -- the `original_description` is there but ideally the algorithm gets away with just using fields immediately on the sortable relation description to keep the up the appearance of the facade.

python/etl/names.py
tvogels01: Appending my original comment ... The problem that I see is that we need to avoid having the same name for different things (and my comment is an example of that confusion). So for me, the base schemas are the list of schema names in order of the config. If we want to have something to easily check against, then a set is great but should be named for what it is used for. Keeping `_base_schemas` and changing the type will get confusing. Since this field is used to detect "managed" schemas, should it then be called just that?
```
        ...
        self._managed_schemas = frozenset()

    @managed_schemas.setter
    def managed_schemas(self, schemas: List) -> None:
        self._managed_schemas = frozenset(schemas)
```

python/etl/names.py
aheyman11: I can't find any instances where we would be comparing staging and non-staging versions of the same table, so I'm going to redefine the hash function as `hash(self.to_tuple())`.

FWIW, the original motivation behind excluding `_base_schemas` from the hash was only that that object was a list and thus non-hashable.

python/etl/names.py
aheyman11: I hadn't realized that we also use base schemas in `TableSelector`s. I've renamed the new concept to `managed_schemas`, which is now a frozenset of strings.

python/etl/relation.py
```diff
@@ -373,25 +374,27 @@ def order_by_dependencies(relation_descriptions):
...
+        unknowns = description.dependencies - known_tables - unmanaged_dependencies
         if unknowns:
             known_unknowns.update(unknowns)
-            has_unknown_dependencies.add(description.identifier)
-            # Drop the unknowns from the list of dependencies so that the loop below doesn't wait for their resolution.
-            description.dependencies = description.dependencies.difference(unknowns)
+            has_unknown_dependencies.add(description.original_description.target_table_name)
+        if unmanaged_dependencies:
+            logger.info("The following dependent relations are not managed by Arthur: %s",
+                        join_with_quotes(unmanaged_dependencies))
```
tvogels01: The `unmanaged_dependencies` here are `List[TableName]`? If so then, it would be nice to unpack their identifiers like in the other occasions when using `join_with_quotes`.
Also, are these "dependent relations"? Aren't these relations that this relation depends on?

python/etl/relation.py
aheyman11: Fixed

python/etl/relation.py
```diff
@@ -373,25 +374,27 @@ def order_by_dependencies(relation_descriptions):
...
-            has_internal_dependencies.add(description.identifier)
+            has_internal_dependencies.add(description.original_description.target_table_name)
         queue.put((1, initial_order, description))
     if has_unknown_dependencies:
         logger.warning("These relations were unknown during dependency ordering: %s",
-                       join_with_quotes(known_unknowns))
+                       join_with_quotes([d.identifier for d in known_unknowns]))
         logger.warning('This caused these relations to have dependencies that are not known: %s',
-                       join_with_quotes(has_unknown_dependencies))
+                       join_with_quotes([d.identifier for d in has_unknown_dependencies]))
```
tvogels01: Should we upgrade `join_with_quotes` to automatically use the identifier?
```python
    if isinstance(names, (set, frozenset)):
        if all(hasattr(name, "identifier") for name in names):
            return ', '.join(sorted("'{}'".format(name.identifier) for name in names))
        else:
            return ', '.join("'{}'".format(name) for name in sorted(names))
    else:
        return ', '.join("'{}'".format(name) for name in names)
```

aheyman11:
Besides the `join_with_quotes` discussion, I *believe* that all other comments have been addressed.

python/etl/names.py
```diff
@@ -115,6 +127,20 @@ def table(self):
...
 
+    @property
+    def managed_schemas(self) -> frozenset:
+        if self._managed_schemas is not None:
+            return self._managed_schemas
+        try:
+            schemas = etl.config.get_dw_config().schemas
+        except AttributeError:
+            raise ETLSystemError("dw_config has not been set!")
+        return frozenset(schema.name for schema in schemas)
```
bhtucker: Does this ever set `self._managed_schemas` to be the result here? Not sure if it's left out or there's some magic I'm missing

PR #62, opened 2017-12-10
[Load fault tolerance] Retry COPY, fall back to prior contents on failure

bhtucker:
On the retry side:
* Add retry support for COPY following the pattern used in extract (assume any error that might be temporary is temporary)
* Add exponential backoff to retry, increase default retry attempts to 3
* In transaction: Don't retry if you're  (even a system error will spoil your transaction, you must give up)

Unclear if retrying COPY helps us or not (it already [claims to retry files when loading from manifests](http://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-multiple-files.html)) but it seems possible. Exponential sleep is known to be useful for extract, as scarce resources (eg, driver memory) may free up between attempts if you wait before retrying.

On the prior contents side:
* No change is usually preferable to total failure (new data may violate uniqueness constraints, but prior data apparently didn't), so try filling a failed relation from prior contents before bubbling error up
* To avoid restoring data referencing out-of-date warehouse-generated row keys, tables with `*_key` columns that aren't generated using the `"identity"` column property cannot be restored.
* Concurrent load will load from prior when data source relations' `extract`s failed or timed out
* Sometimes the failure will occur several relations after the root cause, so developers ought to be able to flag relations for `load_from_prior` so that they can re-load the warehouse with working contents at the root cause relation
* In transaction: Check transaction status and attempt to use prior contents if the transaction is not in an error state (could be fine if problem was a uniqueness constraint or missing SQL file)

### DIFF
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -3,7 +3,15 @@
     "arthur_settings": {
         # If an extract from an upstream source or copy from S3 files fails due to some transient error, retry the extract at most this many times. Zero disables retries
         "extract_retries": 3, 
-        "copy_data_retries": 3
+        "copy_data_retries": 3,
+        # If normal approach to refreshing a materialized relation (CTAS or source) fails, fill the relation with
+        # data already in the warehouse from a prior build of that relation.
+        # Setting determines when Arthur should fallback to prior data:
+        #   "designed" - use (instead of query) only on relations whose design file explicitly sets "loads_from_prior_data"
+        #   "source" - use for source relations where load has failed or concurrent extract failed or timed out
+        #   "non_key" - use for any failed materialized relation other than CTAS with non-"identity" columns named '*_key'
+        #   "all" - use for any failed materialized relation
+        "loads_from_prior_data": "source"
     },
     # Target (Redshift) cluster
     "data_warehouse": {
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -1,8 +1,9 @@
 {
     # General settings controlling how Arthur itself behaves
     "arthur_settings": {
-        # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries
-        "extract_retries": 1
+        # If an extract from an upstream source or copy from S3 files fails due to some transient error, retry the extract at most this many times. Zero disables retries
+        "extract_retries": 3, 
+        "copy_data_retries": 3
     },
     # Target (Redshift) cluster
     "data_warehouse": {
--- a/python/etl/config/table_design.schema
+++ b/python/etl/config/table_design.schema
@@ -79,6 +79,10 @@
             "type" : "string",
             "pattern": "^(CTAS|VIEW|\\w[^.]*[.]\\w[^.]*[.]\\w[^.]*)$"
         },
+        "load_from_prior": {
+            "type": "boolean",
+            "description": "Optional. Indicates an unhealthy table that should retain its last successful contents rather than be updated."
+        },
         "unload_target" : {
             "type": "string",
             "description": "Optional. If the relation is unloadable will specify to which schemas."
--- a/python/etl/config/table_design.schema
+++ b/python/etl/config/table_design.schema
@@ -79,7 +79,7 @@
             "type" : "string",
             "pattern": "^(CTAS|VIEW|\\w[^.]*[.]\\w[^.]*[.]\\w[^.]*)$"
         },
-        "load_from_prior": {
+        "loads_from_prior_data": {
             "type": "boolean",
             "description": "Optional. Indicates an unhealthy table that should retain its last successful contents rather than be updated."
         },
--- a/python/etl/design/redshift.py
+++ b/python/etl/design/redshift.py
@@ -13,7 +13,7 @@ from psycopg2.extensions import connection  # only for type annotation
 
 import etl.db
 from etl.names import join_column_list, TableName
-from etl.errors import TransientETLError, retry
+from etl.errors import TransientETLError
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
--- a/python/etl/design/redshift.py
+++ b/python/etl/design/redshift.py
@@ -13,6 +13,7 @@ from psycopg2.extensions import connection  # only for type annotation
 
 import etl.db
 from etl.names import join_column_list, TableName
+from etl.errors import TransientETLError, retry
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
@@ -189,7 +190,10 @@ def copy_from_uri(conn: connection, table_name: TableName, s3_uri: str, aws_iam_
     else:
         logger.info("Copying data into '%s' from '%s'", table_name.identifier, s3_uri)
         with log_load_error(conn):
-            etl.db.execute(conn, stmt, (s3_uri, credentials))
+            try:
+                etl.db.execute(conn, stmt, (s3_uri, credentials))
+            except psycopg2.InternalError as exc:
+                raise TransientETLError(exc) from exc
             row_count = etl.db.query(conn, "SELECT pg_last_copy_count()")
             logger.info("Copied %d rows into '%s'", row_count[0][0], table_name.identifier)
 
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -59,7 +59,7 @@ import etl.monitor
 import etl.db
 import etl.design.redshift
 import etl.relation
-from etl.config import get_config_int
+from etl.config import get_config_int, get_config_value
 from etl.config.dw import DataWarehouseSchema
 from etl.errors import (ETLRuntimeError, FailedConstraintError, MissingManifestError, RelationDataError,
                         RelationConstructionError, RequiredRelationLoadError, UpdateTableError, retry)
@@ -358,7 +358,7 @@ def copy_data(conn: connection, relation: LoadableRelation, dry_run=False):
     A manifest for the CSV files must be provided -- it is an error if the manifest is missing.
     If the database reports an internal error and we aren't in a transaction, we'll retry.
     """
-    aws_iam_role = str(etl.config.get_config_value("object_store.iam_role"))
+    aws_iam_role = str(get_config_value("object_store.iam_role"))
     s3_uri = "s3://{}/{}".format(relation.bucket_name, relation.manifest_file_name)
 
     if not relation.has_manifest:
@@ -444,15 +444,17 @@ def load_relation_from_prior_data(conn: connection, relation: LoadableRelation, 
     else:
         prior_name = relation.target_table_name.as_backup_table_name()
 
-    key_columns = [column for column in relation.table_design["columns"]
-                   if column["name"].endswith("_key") and not column.get("identity")]
-    if key_columns:
-        raise RelationDataError("Non-identity key columns (%s) may have changed in related tables,"
-                                " reloading table is unsafe.", join_column_list([col['name'] for col in key_columns]))
+    if relation.is_transformation:
+        key_columns = [column for column in relation.table_design["columns"]
+                       if column["name"].endswith("_key") and not column.get("identity")]
+        if key_columns and get_config_value("arthur_settings.loads_from_prior_data") != "all":
+            raise RelationDataError("Non-identity key columns ({}) may have changed in related tables,"
+                                    " reloading table is unsafe.".format(
+                                        join_column_list([col['name'] for col in key_columns])))
 
     logger.info("Loading {:x} with rows from {:s}".format(relation, prior_name))
-    inner_stmt = "SELECT {} FROM {}".format(join_column_list(relation.unquoted_columns), prior_name)
-    insert_from_query(conn, relation, query_stmt=inner_stmt, dry_run=dry_run)
+    prior_data_select_stmt = "SELECT {} FROM {}".format(join_column_list(relation.unquoted_columns), prior_name)
+    insert_from_query(conn, relation, query_stmt=prior_data_select_stmt, dry_run=dry_run)
 
 
 def load_ctas_using_temp_table(conn: connection, relation: LoadableRelation, dry_run=False) -> None:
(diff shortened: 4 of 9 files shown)


python/etl/design/redshift.py
```diff
@@ -13,6 +13,7 @@
 
 import etl.db
 from etl.names import join_column_list, TableName
+from etl.errors import TransientETLError, retry
```
aheyman11: Doesn't look like `retry` gets used in this module?

bhtucker: Ah yes I moved it out a layer

python/etl/load.py
```diff
@@ -427,6 +435,26 @@ def create_missing_dimension_row(columns: List[dict]) -> List[str]:
...
+        prior_name = relation.target_table_name.as_backup_table_name()
+
+    key_columns = [column for column in relation.table_design["columns"]
+                   if column["name"].endswith("_key") and not column.get("identity")]
+    if key_columns:
+        raise RelationDataError("Non-identity key columns (%s) may have changed in related tables,"
+                                " reloading table is unsafe.", join_column_list([col['name'] for col in key_columns]))
+
+    logger.info("Loading {:x} with rows from {:s}".format(relation, prior_name))
+    inner_stmt = "SELECT {} FROM {}".format(join_column_list(relation.unquoted_columns), prior_name)
```
tvogels01: The `inner_` in `inner_stmt` doesn't make much sense here.

python/etl/load.py
```diff
@@ -560,7 +590,9 @@ def update_table(conn: connection, relation: LoadableRelation, dry_run=False) ->
     then it's assumed to be a dimension and a row with missing values (mostly 0, false, etc.) is added as well.
     """
     try:
-        if relation.is_ctas_relation:
+        if from_prior:
```
tvogels01: Are there any safeguards to make sure this isn't called for CTAS?

python/etl/load.py
```diff
@@ -427,6 +435,26 @@ def create_missing_dimension_row(columns: List[dict]) -> List[str]:
...
 
 
+def load_relation_from_prior_data(conn: connection, relation: LoadableRelation, dry_run=False) -> None:
+    """
+    Load relation by selecting contents out of that table's prior position
+    """
+    if relation.use_staging:
+        prior_name = relation.target_table_name.as_standard_table_name()
+    else:
+        prior_name = relation.target_table_name.as_backup_table_name()
```
tvogels01: What happens if the backup table doesn't exist?  What's the error? And would it be clear where that error came from?

python/etl/load.py
```diff
@@ -597,9 +629,21 @@ def build_one_relation(conn: connection, relation: LoadableRelation, dry_run=Fal
             logger.info("Skipping loading data into {:x}".format(relation))
         elif relation.failed:
             logger.info("Bypassing already failed relation {:x}".format(relation))
+        elif relation.load_from_prior:
+            update_table(conn, relation, from_prior=True, dry_run=dry_run)
```
tvogels01: Why are we not verifying constraints when loading prior data?

python/etl/load.py
```diff
@@ -597,9 +629,21 @@ def build_one_relation(conn: connection, relation: LoadableRelation, dry_run=Fal
...
-            verify_constraints(conn, relation, dry_run=dry_run)
+            try:
+                update_table(conn, relation, dry_run=dry_run)
+                verify_constraints(conn, relation, dry_run=dry_run)
+            except ETLRuntimeError as exc:
+                if relation.in_transaction and conn.get_transaction_status() is TRANSACTION_STATUS_INERROR:
+                    logger.warning("Transaction is in error state, cannot attempt to use prior contents")
+                    raise
+                else:
+                    logger.exception("Failed to update relation {:x}".format(relation))
```
tvogels01: This might be nicer as `warning` with `exc_info=True`.  It's not an "error" in the sense that recovery is under way. A warning would highlight that processing continues. So when we check for errors in the logs then that should point to terminal errors (meaning that the data definitely couldn't be loaded).

python/etl/load.py
```diff
@@ -688,8 +733,10 @@ def poll_worker():
                 logger.debug("Poller: %s left to poll, %s ready to load, %s elapsed",
                              to_poll.qsize(), to_load.qsize(), timer.elapsed)
                 if checkpoint_queue_size_cutoff == to_poll.qsize() and timer.elapsed > checkpoint_time_cutoff:
-                    raise ETLRuntimeError(
-                        "No new extracts found in last %s seconds, bailing out" % idle_termination_seconds)
+                    logger.error("No new extracts found in last %s seconds, setting remaining relations to"
+                                 " load from prior and enqueuing" % idle_termination_seconds)
+                    load_remaining_from_prior = True
```
tvogels01: Is it clearly understood by stakeholders what this means? How would we know this happened?

python/etl/names.py
```diff
@@ -127,6 +133,10 @@ def table(self):
     def staging(self):
         return self._staging
 
+    @property
```
tvogels01: I'd prefer to use names that highlight that these are boolean.  So `is_staging` and `is_backup` or something better in this direction.

python/etl/names.py
```diff
@@ -144,7 +154,6 @@ def managed_schemas(self, schema_names: List) -> None:
     def to_tuple(self):
         """
         Return schema name and table name as a handy tuple.
-
```
tvogels01: Please put the empty line between the description and the additional documentation and test back.

The idea for docstrings is to have a short summary, ideally one line. Then there is an empty line.  Then we'll more text to describe the method or class and some doctests.
See [PEP 257](https://www.python.org/dev/peps/pep-0257/) for best practices here.

python/etl/names.py
```diff
@@ -279,6 +282,14 @@ def as_staging_table_name(self):
         tn._staging = True
         return tn
 
+    def as_standard_table_name(self):
+        return TableName(*self.to_tuple())
+
+    def as_backup_table_name(self):
```
tvogels01: This perpetuates a bad pattern. Setting `_backup` and `_staging` should happen in the constructor so that you can easily see by inspecting `__init__` what the read-only fields of the instance are set to.

python/etl/relation.py
```diff
@@ -161,6 +162,14 @@ def is_transformation(self) -> bool:
     def is_unloadable(self) -> bool:
         return "unload_target" in self.table_design
 
+    @property
+    def load_from_prior(self):
```
tvogels01: Same comment as for previous boolean properties.  Would you expect `load_from_prior` to return a boolean?

python/etl/names.py
bhtucker: Yep, incremental change in keeping with the bad pattern. See question of 'position' attribute — could simply default to 'standard' and give a good way to rewrite these bad citizens.

python/etl/names.py
bhtucker: Wrote that for a second and then deleted it! Do you think we should just make an attribute for 'position' or similar instead?

python/etl/load.py
bhtucker: How we'd know is IMO a question for log collection/alerting (lambda hits SNS with warnings/errors?) but the question for stakeholders is a good one. The business consequences of this PR are a big matter for consideration, maybe not in this repo though.

python/etl/load.py
bhtucker: Ah, `exc_info=True` may be better. Wanted to make sure the initial cause of trouble was very visible!

python/etl/load.py
bhtucker: This is definitely a point for discussion — presumably a built table passed its checks when it was made. Do you check the restored rows against the current constraints? My logic was that since this table is or was user facing, it's fair game to restore.

python/etl/load.py
bhtucker: Meant for anything that's materialized (CTAS or source)

python/etl/load.py
bhtucker: The `SELECT` in `INSERT... SELECT` seemed `inner` — but agreed that this is a different 'inner' than the other place this happens. I'll find another name

python/etl/load.py
bhtucker: I've tried to log/use exception blocks carefully to make it clear — when that happens in a fallback case, the original exception was logged, and the new exception will appear in the 'During the handling of the above exception..' context. The inner error would be a psycopg2 programming error, and this would bubble up to an `UpdateTableError` since it's accessed through `update_table`

python/etl/load.py
```diff
@@ -354,20 +356,26 @@ def copy_data(conn: connection, relation: LoadableRelation, dry_run=False):
...
-                            relation.identifier, s3_uri))
+                        relation.identifier, s3_uri))
         else:
             raise MissingManifestError("relation '{}' is missing manifest file '{}'".format(
-                                           relation.identifier, s3_uri))
+                                       relation.identifier, s3_uri))
 
-    etl.design.redshift.copy_from_uri(conn, relation.target_table_name, s3_uri, aws_iam_role,
-                                      need_compupdate=relation.is_missing_encoding, dry_run=dry_run)
+    def _copy_from_s3(attempt_num=0):
```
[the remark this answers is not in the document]
bhtucker: Great idea

python/etl/load.py
```diff
@@ -326,7 +328,7 @@ def grant_access(conn: connection, relation: LoadableRelation, dry_run=False):
     if reader_groups:
         if dry_run:
             logger.info("Dry-run: Skipping granting of select access on {:x} to {}".format(
-                            relation, join_with_quotes(reader_groups)))
```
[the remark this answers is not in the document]
bhtucker: Interesting, this was just response to linter highlighting. Makes sense

python/etl/config/table_design.schema
```diff
@@ -79,6 +79,10 @@
             "type" : "string",
             "pattern": "^(CTAS|VIEW|\\w[^.]*[.]\\w[^.]*[.]\\w[^.]*)$"
         },
+        "load_from_prior": {
```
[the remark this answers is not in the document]
bhtucker: Meant for anything that's materialized (CTAS or source)

python/etl/config/default_settings.yaml
```diff
@@ -1,8 +1,9 @@
 {
     # General settings controlling how Arthur itself behaves
     "arthur_settings": {
-        # If an extract from an upstream source fails due to some transient error, retry the extract at most this many times. Zero disables retries
-        "extract_retries": 1
+        # If an extract from an upstream source or copy from S3 files fails due to some transient error, retry the extract at most this many times. Zero disables retries
+        "extract_retries": 3,
```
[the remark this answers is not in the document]
bhtucker: We know it helps when sqoop can't get the resources to get started — I've never seen an extract run for a long time and then fail (other than when the EMR resources are shut down, at which point, no retries)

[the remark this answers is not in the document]
bhtucker: Did some more research on this: we've had one failed extract that took over 30s in the last 90 days. Sqoop failed with `org.postgresql.util.PSQLException: Connection refused` on both retry attempts. The first attempt failed fast (<5 seconds) but the second attempt took 10 minutes to fail. Since this was for the first relation in a source, this was the busiest time for YARN, so the 10 minute delay is likely waiting for YARN containers to be allocated to the retry. Since the first attempt failed in 5s, I think this attempt also failed quickly once it got resources. I don't think retry attempts would substantially push out the 'Extract' activity time since the only extracts that fail after a while are due to YARN waiting, and there's only YARN waiting when several sources (> 4, in our setup) are still doing work.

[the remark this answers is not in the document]
bhtucker: Adding in development data, we had one other extract failure over 30s which was due to S3 timing: after failing with `No valid CSV files (_SUCCESS is missing)`, the retry attempt starts by deleting the `csv/_SUCCESS` file one second later!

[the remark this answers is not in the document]
tvogels01: Sounds like we could have waited for a few seconds to see whether the sentinel file arrives instead of re-extracting in this case.

python/etl/config/table_design.schema
```diff
@@ -79,6 +79,10 @@
             "type" : "string",
             "pattern": "^(CTAS|VIEW|\\w[^.]*[.]\\w[^.]*[.]\\w[^.]*)$"
         },
+        "loads_from_prior_data": {
```
tvogels01: This feels fragile (is prior data actually available) and also changes concerns of table design. If we think of the table design files outlining the structure of the table and how to (re) create it, then a temporary fix doesn't fit.  My understanding is that we'd have to set this boolean, deploy the table design, run a partial ETL, then not forget to change this back.  This doesn't feel like a step in the right direction for operational ease.

python/etl/load.py
```diff
@@ -51,16 +51,18 @@
...
 from psycopg2.extensions import connection  # only for type annotation
+from psycopg2.extensions import TRANSACTION_STATUS_INERROR
 
 import etl
 import etl.data_warehouse
 import etl.monitor
 import etl.db
 import etl.design.redshift
 import etl.relation
+from etl.config import get_config_int, get_config_value
```
tvogels01: This is an occasion where I'd prefer to stay with the complete name, e.g. `etl.config.get_config_value`, to make it easier to understand that `get_config_value`  actually comes from a  different module and is not implemented here.
(The reason with some other `from .. import` is brevity with clarity, like in case of errors (which we agreed are in only one place) and type checking.)

python/etl/load.py
tvogels01: This scares me.  I feel that this should be discussed how we monitor this and how we distinguish error cases.

python/etl/names.py
tvogels01: I'd say `if name.is_staging` reads nicely compared to `if name.staging` or `if name.position == "staging"` or (worse) `if name.position == TableName.STAGING`. Frankly, if `is_staging` is too short and doesn't read well, then 'in_staging_position` would be better...  Just `staging` or `backup` by themselves don't answer questions which boolean attributes probably should.

bhtucker:
Pulling this apart into separate PRs. Retry copy stands alone cleanly, but even getting 'load from prior' as a fallback for failed extracts requires resolving some of the 'position naming' questions

tvogels01:
Closing this for now. We have pulled some of the code into other PRs and the POC needs to be redone anyways.

PR #63, opened 2017-12-11
Specify columns in copy from S3

aheyman11:
Specify column list in the copy command from S3 to Redshift. This allows static sources to possibly have extra columns (to the right) not loaded by Arthur.

I couldn't add a type annotation to `relation` in `copy_from_uri` because importing `LoadableRelation` would introduce circular dependencies.

### DIFF
--- a/python/etl/design/redshift.py
+++ b/python/etl/design/redshift.py
@@ -166,15 +166,19 @@ def log_load_error(cx):
         raise
 
 
-def copy_from_uri(conn: connection, table_name: TableName, s3_uri: str, aws_iam_role: str,
+def copy_from_uri(conn: connection, relation, s3_uri: str, aws_iam_role: str,
                   need_compupdate=False, dry_run=False) -> None:
     """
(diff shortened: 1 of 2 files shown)


bhtucker:
LGTM

bhtucker:
I'll confess that I argued for passing the relation rather than those two arguments.

What about a base class defined in `relation.py` that both `RelationDescription` and `LoadableRelation` can extend? Even if they aren't extending each other, would be good to show they are the same 'type'

tvogels01:
Well, I'm not writing Java for a reason ;-)  (That being said, having interfaces would be nice at times.)

There are cases in the code already where we're staying away from relation descriptions, e.g when they are built. Having the pieces (table name and columns) matches this pattern. None of the modules inside `etl.design` need relation descriptions.  Why should `etl.design.redshift`?

tvogels01:
We should merge this along with PR #61 into a new version (v1.6.1?)

PR #70, opened 2017-12-20
[Load fault tolerance] Retry COPY

bhtucker:
We see transient-looking errors from S3 data files not being found/readable during COPY commands. We're optimistic that retrying might succeed.

User facing changes:
* Retry COPY from S3 whenever a Redshift InternalServerError is encountered

Internal changes:
* Add exponential backoff to retries

Note that this has a difference in monitor output from extract, where the entire monitored step is retried, with monitor logging of attempts. Since this retry is for one part of a monitored step but much of the step won't be retried, I don't think it makes sense to expect these retries to look the same to the monitor.

(HA-734)

### DIFF
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -1,8 +1,9 @@
 {
     # General settings controlling how Arthur itself behaves
     "arthur_settings": {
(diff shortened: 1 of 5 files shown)


tvogels01:
LGTM (but I'd like to wait until the new year to test and merge)

Issue #72, opened 2017-12-20
Can't identify final extract failures in monitor payloads


bhtucker:
In terms of output what is recorded by the monitor, failures that will be retried look like final failures.

This poses a problem for concurrent load — load thinks the first failure it sees is final. Since the extract job might have different settings than load, you can't know based on the `attempt_num` if this is the final one.

tvogels01:
I'm closing this for now.  The retries are no longer recorded in the events table so skipping past to the last event doesn't matter. Unfortunately we now have also lost the opportunity to see in the events table whether retries have occurred.

PR #78, opened 2017-12-22
Add an event to distinguish failures that are not final

bhtucker:
A proposed implementation to resolve https://github.com/harrystech/arthur-redshift-etl/issues/72

User facing change:
* Monitor payloads' "event" field may now have a new value, `'fail_before_retry'`, that indicates a step attempt ended in error but will be retried

Internal changes:
* Concurrent load event polling now looks for only the events it can handle ('fail' and 'finish' events) rather than 'everything but start'

(HA-733)

### DIFF
--- a/python/etl/extract/extractor.py
+++ b/python/etl/extract/extractor.py
@@ -12,7 +12,7 @@ from typing import Dict, List, Set
 import etl.monitor
 import etl.s3
 import etl.db
-from etl.config import get_config_int
+import etl.config
 from etl.config.dw import DataWarehouseSchema
 from etl.errors import (
     DataExtractError,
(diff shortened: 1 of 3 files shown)


python/etl/extract/extractor.py
```diff
@@ -77,6 +77,7 @@ def extract_source(self, source: DataWarehouseSchema,
         with Timer() as timer:
             for i, relation in enumerate(relations):
                 try:
+                    retries = get_config_int("arthur_settings.extract_retries")
```
tvogels01: This is duplicate code (the old line still exists below the def) ... Please move this lookup of a runtime config outside the loop (and have it only once ;-)

python/etl/monitor.py
```diff
@@ -47,6 +47,7 @@
 STEP_START = "start"
 STEP_FINISH = "finish"
 STEP_FAIL = "fail"
+STEP_ATTEMPT_FAIL = "attempt_fail"
```
tvogels01: Maybe "fail_before_retry"?

tvogels01:
I think we get ourselves into trouble in load.py. When we receive a "fail before retry" event, then we'd try to load it here: https://github.com/harrystech/arthur-redshift-etl/blob/next/python/etl/load.py#L720

We can either keep the current semantics and treat fail and "fail before retry" the same way. Or simply ignore for now the "fail before retry" events in load.

bhtucker:
Indeed — added a commit to filter explicitly for the events that we handle.

PR #88, opened 2018-01-10
Optionally check for recent successful extracts before running an update

bhtucker:
Attempt to solve https://github.com/harrystech/arthur-redshift-etl/issues/89 by checking for recent extracts during `update` and de-coupling these Data Pipeline Activities.


tvogels01:
Please add an issue as well so that we capture a bit more what specifically we're trying to solve here and what we've discussed as alternatives or follow-up solutions.

bhtucker:
Re: `join_with_quotes`, this seemed like the right time to pick up a suggestion from an earlier discussion that `etl.errors` could import (exclusively) `etl.text`. Seems natural enough, happy to switch back if there are other consequences I'm missing

PR #92, opened 2018-02-15
Move error logging within retry

tvogels01:
Logging needs to be around the execute call s.t. we can see why Redshift rejected
the load based on stl load errors.

### DIFF
--- a/python/etl/design/redshift.py
+++ b/python/etl/design/redshift.py
@@ -193,13 +193,13 @@ def copy_from_uri(conn: connection, table_name: TableName, column_list: List[str
         etl.db.skip_query(conn, stmt, (s3_uri, credentials))
     else:
         logger.info("Copying data into '%s' from '%s'", table_name.identifier, s3_uri)
-        with log_load_error(conn):
-            try:
+        try:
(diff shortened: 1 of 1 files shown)


python/etl/design/redshift.py
```diff
@@ -193,13 +193,13 @@ def copy_from_uri(conn: connection, table_name: TableName, column_list: List[str
...
         logger.info("Copying data into '%s' from '%s'", table_name.identifier, s3_uri)
-        with log_load_error(conn):
-            try:
+        try:
+            with log_load_error(conn):
                 etl.db.execute(conn, stmt, (s3_uri, credentials))
-            except psycopg2.InternalError as exc:
-                raise TransientETLError(exc) from exc
             row_count = etl.db.query(conn, "SELECT pg_last_copy_count()")
             logger.info("Copied %d rows into '%s'", row_count[0][0], table_name.identifier)
```
michael-john-fu: f strings?

tvogels01: Ideally ... yes.  We need to update your EMR setup to use new Python.

PR #110, opened 2019-03-11
Updates for v1.8.0 to play with Docker

tvogels01:
With these changes, the docker image builds nicely and running `arthur.py` inside a running container works.

In particular:
```
arthur.py settings object_store.s3.*
arthur.py ping
arthur.py sync --deploy
```

### DIFF
--- a/bin/docker_run.sh
+++ b/bin/docker_run.sh
@@ -1,34 +1,40 @@
 #!/bin/bash
+
 set -eu
 
 show_usage_and_exit () {
     cat <<EOF
-Usage: `basename $0` [-p aws_profile] <config_dir> [<target_env>]
+
+Usage: `basename $0` [-p aws_profile] [-t tag] [<config_dir> [<target_env>]]
 
 This will drop you into a shell in a Docker container with Arthur installed and
 configured to use <config_dir>.
+
+The <config_dir> default to \$DATA_WAREHOUSE_CONFIG.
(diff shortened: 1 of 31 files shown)


bin/docker_run.sh
```diff
@@ -36,11 +38,30 @@ while getopts "hp:" opt; do
...
+else
+    profile_arg=""
+fi
 
-docker run --rm -it -v "$PWD":/data-warehouse -v ~/.aws:/root/.aws -e DATA_WAREHOUSE_CONFIG=/data-warehouse/$config_dir -e USER=$target_env $profile harrystech/arthur
+set -x
+docker run --rm -it -v "$PWD":/data-warehouse -v ~/.aws:/root/.aws \
+    -e DATA_WAREHOUSE_CONFIG=/data-warehouse/$config_dir \
+    -e "ARTHUR_DEFAULT_PREFIX=$target_env" -e "USER=$target_env" \
+    $profile_arg harrystech/arthur
```
blohr-hs: Actually, looking at this now, I wonder if it makes sense to remove `harrystech` from the image name, since this is technically a general open source project.

tvogels01: Agreed. Do you want to just drop it or use a parameter?
(No harm in somebody else creating that image but it's indeed weird.)

blohr-hs:
LGTM, :shipit:

PR #128, opened 2019-04-17
Add PagerNotification topic for Validation Pipeline on failure

marwamc-hs:
Made changes to create a "validation-page" topic.
In validation pipeline template, added onFail block that will send a message to this topic.

Once this topic is created in production, will add a subscription that pings PagerDuty.

### DIFF
--- a/python/etl/templates/validation_pipeline.json
+++ b/python/etl/templates/validation_pipeline.json
@@ -39,6 +39,13 @@
             "message": "Failed step #{node.name} at #{node.@actualEndTime}\\nPipelineId: #{node.@pipelineId}\\nSphere: #{node.@sphere}\\nCancellation reason: #{node.cancellationReason}\\nLog directory: #{node.pipelineLogUri}#{node.@pipelineId}\\n\\nError stacktrace: #{node.errorStackTrace}"
         },
         {
+            "id": "PagerNotification",
+            "type": "SnsAlarm",
+            "topicArn": "arn:aws:sns:${resources.VPC.region}:${resources.VPC.account}:${resource_prefix}-validation-page",
+            "subject": "ETL Rebuild Failure: ${object_store.s3.prefix} at #{node.@scheduledStartTime}",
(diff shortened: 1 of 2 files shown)


python/etl/templates/validation_pipeline.json
```diff
@@ -126,7 +133,11 @@
             "parent": { "ref": "ArthurCommandParent" },
             "command": "/tmp/redshift_etl/venv/bin/arthur.py --config /tmp/redshift_etl/config/ validate --prolix --remote-files --prefix ${object_store.s3.prefix}/validation --keep-going --skip-sources-check",
             "dependsOn": { "ref": "ArthurLoad" },
-            "onSuccess": {"ref": "SuccessNotification"}
+            "onSuccess": {"ref": "SuccessNotification"},
+            "onFail": [
+                { "ref": "FailureNotification" },
+                { "ref": "PagerNotification" }
```
blohr-hs: Given that `FailureNotification` is already an SNS topic that only fires on pipeline failure, why do we need a separate `PagerNotification` topic? Seemingly we could subscribe PagerDuty to the existing `FailureNotification ` topic to achieve the desired result.

marwamc-hs: Currently, both FailureNotification and SuccessNotification have the SAME topic arn.
The PagerNotification has a separate topic arn. This the one that will then push to PagerDuty.
If we attempt to push to pagerduty from the FailureNotification topic arn, we will receive HIGH alerts for both successes and failures.

tvogels01: The small annoyance here is that users now get two emails when their validation pipeline fails.

The `-validation` topic receives notifications for success and failure and so the user receives email for both, success and failure.

The `-validation-page` topic receives notifications for failure and so the user receives an additional (and almost identical) email about the failure.

If this gets annoying, we can unsubscribe from the `-validation-page` topic.

marwamc-hs: currently, no users are subscribed to the `-validation-page` topic. The intention is that the only Subscriber to this topic will be the PagerDuty email.

tvogels01: Remember that `sns_subscribe.sh` will subscribe a user to all of their topics automatically. So this *will* subscribe users to both, `-validation` and `-validation-page`.

Also, in order to create the topic, you'll have to run `sns_subscribe.sh` for our group email. (Remember this repo is public, so let's not add specifics :-) )

PR #131, opened 2019-05-07
Add row count to output of "query_events"

tvogels01:
Goal is to be able to compare row counts of tables between separate ETL runs.

First, look up the ETLs in the current environment with:
```
arthur.py query_events -q
```
Let's say the output is:
```
etl_id           | step    | timestamp
-----------------+---------+--------------------
C92AC14C4AD8442C | upgrade | 2019-05-05T15:47:11
4EA4FC6494CE496B | upgrade | 2019-05-05T20:55:28
```

If we now fetch the event information, it would create false-positives in comparisons since the time information is included. Consider:
```
# arthur.py query_events -q C92AC14C4AD8442C
target                           | step    | event  | timestamp           | elapsed | rowcount
---------------------------------+---------+--------+---------------------+---------+---------
redshift_experiments.dimensions  | upgrade | finish | 2019-05-05T15:47:56 |  42.85  |  98765
redshift_experiments.facts       | upgrade | finish | 2019-05-05T15:48:43 |  46.80  |  7654321
(2 rows)
```

So compare two runs by selecting only `rowcount` in the output. For example:
```
# arthur.py query_events -q C92AC14C4AD8442C --column rowcount
target                          | event  | rowcount
--------------------------------+--------+---------
redshift_experiments.dimensions | finish | 98765
redshift_experiments.facts      | finish | 7654321
(2 rows)
# arthur.py query_events -q 4EA4FC6494CE496B --column rowcount
target                          | event  | rowcount
--------------------------------+--------+---------
redshift_experiments.dimensions | finish | 44765
redshift_experiments.facts      | finish | 7654321
(2 row2)
```
(In this case ... something changed in the second run with the dimensions ... go fix it :-) )

Overall, we'll always show the `target` and `event` columns.

### DIFF
--- a/python/etl/monitor.py
+++ b/python/etl/monitor.py
@@ -543,14 +543,40 @@ def start_monitors(environment):
 
 
 def _format_output_column(key: str, value: str) -> str:
-    if key == "timestamp":
+    if value is None:
+        return '---'
+    elif key == "timestamp":
         # Make timestamp readable by turning epoch seconds into a date.
         return datetime.utcfromtimestamp(float(value)).replace(microsecond=0).isoformat()
     elif key == "elapsed":
         # Reduce number of decimals to 2.
         return '{:6.2f}'.format(float(value))
+    elif key == "rowcount":
+        return '{:9d}'.format(int(value))
     else:
         return value
+
+
+def _flatten_scan_result(result: dict) -> dict:
+    """Remove type annotation from a scan result.
+
+    Careful, this only works for a specific subset of types that DynamoDB may send back.
+
+    >>> result = _flatten_scan_result({'step': {'S': 'load'}, 'extra': {'M': {'rowcount': {'N': '100'}}}})
+    >>> sorted(result)
+    ['extra', 'step']
+    >>> result['extra']['rowcount']
+    '100'
+    """
+    flat = {}
+    for key, value in result.items():
+        if 'S' in value:
+            flat[key] = value['S']
+        elif 'N' in value:
+            flat[key] = value['N']
+        elif 'M' in value:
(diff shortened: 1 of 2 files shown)


python/etl/monitor.py
```diff
@@ -590,13 +616,24 @@ def query_for_etl_ids(hours_ago=0, days_ago=0) -> None:
...
-    keys = ["target", "step", "event", "timestamp", "elapsed"]
+    all_keys = ["target", "step", "event", "timestamp", "elapsed", "rowcount"]
+    if comma_separated_columns:
+        selected_columns = comma_separated_columns.split(',')
+        invalid_columns = [key for key in selected_columns if key not in all_keys]
+        if invalid_columns:
+            raise InvalidArgumentError("invalid column(s): {}".format(','.join(invalid_columns)))
+        # We will always select "target" and "event" to have a meaningful output.
+        selected_columns = frozenset(selected_columns).union(["target", "event"])
+        keys = [key for key in all_keys if key in selected_columns]
```
blohr-hs: `selected_columns` must be a subset of `all_keys` already, due to the validation, so is there a reason to not just set `keys` to `selected_columns`? It looks like `keys` should just be a set anyway.

blohr-hs:
Any chance you'd be willing to write up a short runbook/playbook on this use case, exploiting this new feature?

python/etl/monitor.py
tvogels01: The reason for creating `keys` is to keep the order of columns stable. `selected_columns` is derived from the command line argument. By tying `keys` to the order of `all_keys`, the column order will be the same no matter which subset is selected (`--columns rowcount,elapsed` vs. `--columns elapsed,rowcount`).

PR #138, opened 2019-10-30
Pin the mypy version

thomas-vogels:
Leaving the mypy version to trail upstream causes some validation issues right now in the type checker. This pins the version to 0.670.


tvogels01:
Oops. Used wrong github account.

PR #143, opened 2019-11-25
Show table execution level

tvogels01:
The output for `show_downstream_dependents` now also shows the "execution level" which starts from 1 for sources (or independent views and tables) and from then on is the max of any level of dependencies plus 1. All tables with the same execution level could be executed in parallel and all would work out in the right order.

The implementation is slightly ugly in that the level is annotated in the original relation description. That's a (design) problem for another day. (On the bright side, I did some code-cleanup around ordering relations in execution order.)

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -1084,7 +1084,7 @@ def show_downstream_dependents(relations: List[RelationDescription], selector: T
                      " flag={flag:9s}"
                      " is_required={relation.is_required}")
 
-    back_reference = {}  # type: Dict[TableName, RelationDescription]
+    relation_map = {relation.target_table_name: relation for relation in selected_relations}
     for i, relation in enumerate(selected_relations):
         if relation.identifier in selected:
             flag = "selected"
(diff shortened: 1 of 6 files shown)


python/etl/load.py
```diff
@@ -1080,8 +1080,11 @@ def show_downstream_dependents(relations: List[RelationDescription], selector: T
                 len(selected), len(immediate), len(selected_relations) - len(selected) - len(immediate))
 
     max_len = max(len(relation.identifier) for relation in selected_relations)
-    line_template = ("{relation.identifier:{width}s}"
-                     " # index={index:4d}, flag={flag:.9s}, kind={relation.kind}, is_required={relation.is_required}")
+    line_template = ("{relation.identifier:{width}s} # kind={relation.kind} index={index:4d} level={relation.level:3d}"
+                     " flag={flag:9s}"
+                     " is_required={relation.is_required}")
+
+    back_reference = {}  # type: Dict[TableName, RelationDescription]
```
soumyadsanyal: (Very non-blocking) Looking at `back_reference` (before skipping ahead to its use below) makes me think of cycle detection -- would it make sense to call this map something like `tablename_relation_map` or `relation_by_tablename`?

soumyadsanyal: E.g.: Seems like a similar use to https://github.com/harrystech/arthur-redshift-etl/pull/143/files#diff-4b4b74e85c91c2422250792d1a1052ffR417 ?

python/etl/relation.py
```diff
@@ -401,35 +386,84 @@ def order_by_dependencies(relation_descriptions):
...
+    queue = PriorityQueue()  # type: PriorityQueue[Tuple[int, int, SortableRelationDescription]]
+    for initial_order, description in enumerate(descriptions):
+        queue.put((1, initial_order, description))
+
     table_map = {description.target_table_name: description for description in descriptions}
     latest = 0
     while not queue.empty():
         minimum, tie_breaker, description = queue.get()
+
         if minimum > 2 * nr_tables:
```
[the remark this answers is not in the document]
tvogels01: Nope. I suspect that this is left over from a time when we looped over all relations that didn't depend on something in `pg_catalog` and then looped over those that did. `minimum` should never grow larger than `nr_tables` because we'd have holes in the orders otherwise.

python/etl/load.py
tvogels01: That's now `relation_map` in both places.

soumyadsanyal:
@tvogels01 Approved, thank you! The only think I wanted to check was some heads down research on the execution model for BFS -- but that's just a note to self, and is not blocking.

PR #150, opened 2020-02-08
Unload with column types

tvogels01:
This adds more information about a relation's columns into our `columns.yaml` file.

```
arthur.py unload example.test
```
results into a new object `columns_with_types` in `columns.yaml`:
```
{
    "columns": [
        "id",
        "an_int",
        "a_double",
        "a_numeric",
        "a_boolean",
        "a_text",
        "a_date",
        "a_timestamp"
    ],
    "columns_with_types": [
        {
            "name": "id",
            "type": "int"
        },
        {
            "name": "an_int",
            "type": "int"
        },
        {
            "name": "a_double",
            "type": "double"
        },
        {
            "name": "a_numeric",
            "type": "string"
        },
        {
            "name": "a_boolean",
            "type": "boolean"
        },
        {
            "name": "a_text",
            "type": "string"
        },
        {
            "name": "a_date",
            "type": "string"
        },
        {
            "name": "a_timestamp",
            "type": "string"
        }
    ]
}
```

The motivation here is that we can now more easily post-process the files, say load them into another (potentially non-Postgres) database.


tvogels01:
@soumyadsanyal I'm thinking that an alternative would be to read the manifest file, merge it in here. This way, downstream apps that use the "unloaded" data have to read only one meta file. That file would provide the list of files and the description of the columns.

If we don't want to do that, then an alternative approach would be to add a new object in `columns.yaml`, say `columns_with_types` and not write an additional file.

soumyadsanyal:
> This adds more information about a relation's columns into our `columns.yaml` file.
> 
> ```
> arthur.py unload example.test
> ```
> 
> results into a new object `columns_with_types` in `columns.yaml`:
> 
> ```
> {
>     "columns": [
>         "id",
>         "an_int",
>         "a_double",
>         "a_numeric",
>         "a_boolean",
>         "a_text",
>         "a_date",
>         "a_timestamp"
>     ],
>     "columns_with_types": [
>         {
>             "name": "id",
>             "type": "int"
>         },
>         {
>             "name": "an_int",
>             "type": "int"
>         },
>         {
>             "name": "a_double",
>             "type": "double"
>         },
>         {
>             "name": "a_numeric",
>             "type": "string"
>         },
>         {
>             "name": "a_boolean",
>             "type": "boolean"
>         },
>         {
>             "name": "a_text",
>             "type": "string"
>         },
>         {
>             "name": "a_date",
>             "type": "string"
>         },
>         {
>             "name": "a_timestamp",
>             "type": "string"
>         }
>     ]
> }
> ```
> 
> The motivation here is that we can now more easily post-process the files, say load them into another (potentially non-Postgres) database.

Something that comes to mind here is that the `columns.yaml` file also makes it possible to record additional metadata about the table that could be used for data quality checks.

E.g.: max / min / range of each column of a totally orderable datatype; means, medians, modes, ...

soumyadsanyal:
@tvogels01 LGTM.

QQ: Did you want to include this metadata in the manifest file? That seems potentially quite useful, but I may be missing context you've developed since our last conversation on this.

tvogels01:
@soumyadsanyal The columns file cannot be included into the manifest so that the interface with Redshift stays the same. (Well, we could test what happens with Redshift if we add a new item in the manifest.) 

As for data quality checks, I'm not sure I'd stuff that into the same file. There are files related to structure and there are files related to runtime (in my head). We could "just" start logging values and log collection would pick up metrics?

PR #153, opened 2020-02-15
Add isort configuration, sort imports of Python files

tvogels01:
Today I've learned that I'm not as good at keeping lists sorted alphabetically as I thought I was.
This just adds the configuration for `isort`. "Enforcement" of keeping import lists sorted will follow in a separate PR. Until then, we'll continue the honor system. (That is, you should run `isort` on all files that you modify before opening a PR.)

To validate:
```
isort --recursive --check-only python
```


pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.isort]
+default_section = "THIRDPARTY"
+include_trailing_comma = true
+force_grid_wrap = 0
+known_first_party = ["etl"]
+line_length = 120
```
marwamc-hs: 100? maybe not here...?

tvogels01: Indeed: not yet :-)

PR #154, opened 2020-02-15
Add "black" to project

tvogels01:
This adds settings for [black](https://github.com/psf/black) with the original line length of 120,  then reformats our Python code.

Goal here is to get to a consistent Python formatting style.  A future PR may reduce the line length to 100 to make PRs easier to review.

To validate this PR (or check future commits):
```
black --check --diff python
```


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [pycodestyle]
-# E252 -- Many of our type annotations are flagged.
-# W504 -- We prefer to break after the binary operator.
-ignore = E252,W504
+# Seetings are compatible with black, see https://github.com/psf/black/issues/354
+ignore = E203,W503
+# This is the current setting for this project. We may want to think to drop to 88.
```
blohr-hs: I thought you wanted line length to be 100?

pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
+include = '\.pyi?$'
+line-length = 120
```
marwamc-hs: 100?

tvogels01: Yes, that's the goal. But it would have required quite a bit of (unscheduled) work to update the code base. I'm more inclined to move whatever module we are working on to the new line length. You know, have the code in a better place when you leave.

setup.cfg
tvogels01: Thanks! Fixed.

PR #156, opened 2020-03-19
Replace Cronut with Healthchecks.io

blohr-hs:
Adds a new script to ping the health check service, and updates the relevant data pipeline templates to call it. Uses a new env var, `HEALTH_CHECKS_URI`, which it expects to find in the `credentials.sh` file like normal.


bin/bootstrap.sh
```diff
@@ -64,7 +64,7 @@ cd "$PROJ_TEMP"
 # Download code to all nodes, this includes Python code and its requirements.txt
 aws s3 cp --only-show-errors --recursive "s3://$BUCKET_NAME/$ENVIRONMENT/jars/" ./jars/
 aws s3 cp --only-show-errors --recursive \
-    --exclude '*' --include bootstrap.sh --include ping_cronut.sh --include sync_env.sh \
+    --exclude '*' --include bootstrap.sh --include send_health_check.sh --include sync_env.sh \
```
tvogels01: Overall my observation is that "send health check" sounds strange. I'd prefer `send_health_check_ping` but can live with what is. (For me "health checks" are inbound, "health check pings" are outbound.)

bin/send_health_check.sh
```diff
@@ -0,0 +1,51 @@
...
+#   HEALTH_CHECK_URI - The bare URI itself, when the job completes
+#   HEALTH_CHECK_URI/start - To indicate the job has started
+#   HEALTH_CHECK_URI/fail - To explicitly indicate job failure
+
+if [[ "$1" == "-h" || ("$1" != "" && "$1" != "start" && "$1" != "fail") ]]; then
+    echo "Usage: $0 [ start | fail ]"
+    exit 1
+fi
+
+set -e
```
tvogels01: I'd prefer that we move to the verbose versions, that is `set -o errexit` etc., if not part of this PR, then in a future PR.

bin/send_health_check.sh
```diff
@@ -0,0 +1,51 @@
...
+    source "$FILENAME"
+else
+    echo "Failed to find '$FILENAME'"
+fi
+if [ -z "$HEALTH_CHECK_URI" ]; then
+    echo "Failed to find value in environment variable HEALTH_CHECK_URI, health checks disabled"
+    exit 0
+fi
+
+# Status should be only one of: "", "start", or "fail"
```
tvogels01: Wouldn't being paranoid here help us to make sure this holds? Running the assertion in code, documents and verifies what's expected.

```diff
diff --git a/bin/send_health_check.sh b/bin/send_health_check.sh
index 883dea6..d18d4dc 100755
--- a/bin/send_health_check.sh
+++ b/bin/send_health_check.sh
@@ -38,14 +38,20 @@ if [ -z "$HEALTH_CHECK_URI" ]; then
     exit 0
 fi
 
-# Status should be only one of: "", "start", or "fail"
-status="$1"
-if [ -n "$status" ]; then
-    HEALTH_CHECK_URI="$HEALTH_CHECK_URI/$status"
-fi
+case "$1" in
+    start|fail)
+        HEALTH_CHECK_URI="$HEALTH_CHECK_URI/$1"
+        ;;
+    "")
+        # Keep URI as-is
+        ;;
+    *)
+        echo "Unexpected arg. Should be missing or 'start' or 'fail'."
+        exit 1
+esac
```

etc/cronut_public_ids.sh_template
```diff
@@ -1,9 +0,0 @@
-# You need to add the Ping IDs from Cronut here.
```
tvogels01: Please add a new template file that demonstrates use of `HEALTH_CHECK_URI`

E.g.
```diff
diff --git a/etc/credentials.sh.template b/etc/credentials.sh.template
new file mode 100644
index 0000000..0859bcc
--- /dev/null
+++ b/etc/credentials.sh.template
@@ -0,0 +1,9 @@
+# Add list of upstream data sources here:
+YOUR_APPLICATION_PRODUCTION=postgres://user:password@host:port/dbname?sslmode=require
+
+# Add connection strings used by arthur to connect to redshift as admin or ETL user:
+DATA_WAREHOUSE_ADMIN=postgres://user:password@polaris.dev.harrys.systems:5439/dev?sslmode=require
+DATA_WAREHOUSE_ETL=postgres://user:password@polaris.dev.harrys.systems:5439/development?sslmode=require
+
+# Add URL to send health check pings to
+HEALTH_CHECK_URI=http://example-health-checker.com/token
```

python/etl/templates/rebuild_pipeline.json
```diff
@@ -161,11 +161,11 @@
...
         },
         {
-            "id": "PingCronutAfterEtl",
-            "name": "Ping Cronut After ETL (EC2)",
+            "id": "SendHealthCheckAfterEtl",
+            "name": "Send Health Check After ETL (EC2)",
             "type": "ShellCommandActivity",
             "parent": { "ref": "ShellCommandParent" },
-            "command": "bash /tmp/redshift_etl/bin/ping_cronut.sh PING_AFTER_ETL",
+            "command": "bash /tmp/redshift_etl/bin/send_health_check.sh",
```
tvogels01: Q: Should we add an explicit ping for `fail` if the steps fail?

bin/bootstrap.sh
blohr-hs: Noted, but it doesn't sound like it's worth changing.

bin/send_health_check.sh
blohr-hs: Will fix

bin/send_health_check.sh
blohr-hs: The beginning of the script already does check this on line 22. But the `case` version is a lot prettier, I'll try to use that.

etc/cronut_public_ids.sh_template
blohr-hs: OK

python/etl/templates/rebuild_pipeline.json
blohr-hs: That's up to you. I think it should have separate SNS topics for success and failure, instead of multiplexing them both on the same `status` topic. If it were set up that way, I could have done this whole integration without touching any shell scripts or pipeline definitions (except to delete cronut references) via SNS subscriptions.

bin/send_health_check.sh
blohr-hs: After putting it in, it seemed too redundant with line 22, and not worth restructuring the logic to avoid using `HEALTH_CHECKS_URI` before checking, so I decided to just leave it as is.

python/etl/templates/rebuild_pipeline.json
tvogels01: I don't think that there was a way of setting up cronut with SNS alone given it's curious API.

Ok, we'll leave it as tech debt to move from scripts on the host to SNS.

PR #157, opened 2020-03-20
Make timeout for rebuild pipeline configurable

blohr-hs:
Updated all of the pipeline templates so that the timeout is now a parameter instead of being hard-coded. The parameters simply default to the previously hard-coded values.

Only the `install_rebuild_pipeline.sh` script was updated to allow the user to set the timeout parameter however; that's the only one we need right now, and it was much simpler to add the extra parameter to that script than several of the others.

Depends on #156

### DIFF
--- a/python/etl/templates/validation_pipeline.json
+++ b/python/etl/templates/validation_pipeline.json
@@ -49,7 +49,7 @@
             "id": "ResourceParent",
             "keyPair": "${resources.key_name}",
             "subnetId": "${resources.VPC.public_subnet}",
-            "terminateAfter": "2 Hours"
+            "terminateAfter": "#{myTimeout} Hours"
         },
         {
             "id": "ArthurDriverEC2Resource",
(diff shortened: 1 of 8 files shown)


python/etl/templates/validation_pipeline.json
```diff
@@ -155,10 +155,20 @@
             "description": "Number of occurrences for this pipeline",
             "watermark": "1000",
             "helpText": "How often should the pipeline schedule be repeated?"
+        },
+        {
+            "id": "myTimeout",
+            "type": "Integer",
+            "optional": "true",
+            "description": "How many hours to allow the pipeline to run before terminating it",
+            "watermark": "6",
```
tvogels01: The default value should be `1`.

python/scripts/install_rebuild_pipeline.sh
```diff
@@ -1,11 +1,15 @@
 #!/usr/bin/env bash
 
-if [[ $# -ne 3 || "$1" = "-h" ]]; then
-    echo "Usage: `basename $0` <environment> <startdatetime> <occurrences>"
+if [[ $# -lt 3 || "$1" = "-h" ]]; then
```
tvogels01: The other side is missing (`$# -gt 4`)

setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name="redshift_etl",
-    version="1.14.0",
+    version="1.15.0",
```
tvogels01: We'll release the PRs together. We only need to bump setup.py once per release.

PR #159, opened 2020-03-26
Remove pyproject.toml file from dist file

tvogels01:
Looks like there's a bug affecting `pip3` when there's a `pyproject.toml` file at the toplevel of a project. As a "hotfix", this PR moves this file into `etc/` which then means it actually doesn't get copied into the tar ball when running `sdist`.

This also adds documentation for using the `etc/pyproject.toml` file with `black` and `isort`.


soumyadsanyal:
> will go with faith here.

Доверя́й, но проверя́й

PR #166, opened 2020-04-07
Mount source volume

tvogels01:
Adding the current directory when using `bin/run_arthur.sh` so that one can develop on the ETL code inside the Docker container.

### DIFF
--- a/bin/docker_run.sh
+++ b/bin/docker_run.sh
@@ -116,6 +116,7 @@ case "$action" in
         docker run --rm --interactive --tty \
             --publish 8086:8086/tcp \
             --volume "$data_warehouse_path":/data-warehouse \
+            --volume `pwd`:/arthur-redshift-etl \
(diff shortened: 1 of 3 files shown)


blohr-hs:
You could automate the setup part with a script that runs in the container and does the installation prior to dropping the user to the shell. That way you run one command, and land in a read-to-go environment.

bin/docker_run.sh
```diff
@@ -115,10 +116,12 @@ case "$action" in
         docker run --rm --interactive --tty \
             --publish 8086:8086/tcp \
             --volume "$data_warehouse_path":/data-warehouse \
+            --volume `pwd`:/arthur-redshift-etl \
```
blohr-hs, line 119: This looks like it's being mounted twice; is that needed?

tvogels01: Thanks for catching that. Bad merge mojo.

PR #169, opened 2020-04-08
Build Arthur so that latest version is used for deployment

tvogels01:
When deploying arthur, it was assumed that the latest version had been built. This now simply builds arthur before deploying it so that the release info and build image are up to date.


bin/docker_run.sh
```diff
@@ -141,6 +141,9 @@ case "$action" in
         ;;
     upload)
         set -o xtrace
+        bin/release_version.sh
+        docker build --tag "arthur:$tag" .
```
marwamc-hs, line 145: will save us some painful moments :)

PR #171, opened 2020-04-09
Use schema name as default source-schema name

tvogels01:
I'm revisiting our filename patterns in the context of supporting file formats other than CSV. This made me wonder whether we still should keep that awkward duplication of the schema name for transformations. So instead of:
```
schemas/dw/dw-fact_users.sql
schemas/dw/dw-fact_users.yaml
```
we'd just have
```
schemas/dw/fact_users.sql
schemas/dw/fact_users.yaml
```

How to test -- first on `next`
* Change some files that are for transformations by dropping the second schema name. For the example above:
```
mv schemas/dw/dw-fact_order.sql schemas/dw/fact_order.sql
mv schemas/dw/dw-fact_order.yaml schemas/dw/fact_order.yaml
```
* While on next, run the `ls` command _after building the image_:
```
arthur.py ls dw.fact_order
```
and watch the error pop up:
```
Bailing out: FileNotFoundError: Found no matching files in './schemas' for '['dw.fact_order']'
```

How to test -- then on this branch with the same change of file names from above:
* While on this branch, run the `ls` command _after building the image_:
```
arthur.py ls dw.fact_order
```
and observe the simple output
```
Schema: 'dw'
    Table: 'dw.fact_order'
        schemas/dw/fact_order.yaml
        schemas/dw/fact_order.sql
```

For more testing, you should also be able to do:
```
arthur.py sync dw.fact_order
arthur.py upgrade --dry-run dw.fact_order
```

### DIFF
--- a/python/etl/file_sets.py
+++ b/python/etl/file_sets.py
@@ -231,10 +231,10 @@ def _find_matching_files_from(iterable, pattern, return_success_file=False):
     """
     file_names_re = re.compile(
         r"""(?:^schemas|/schemas|^data|/data)
-                                   /(?P<source_name>\w+)
-                                   /(?P<schema_name>\w+)-(?P<table_name>\w+)
-                                   (?:(?P<file_ext>.yaml|.sql|.manifest|/csv/(:?part-.*(:?\.gz)?|_SUCCESS)))$
-                               """,
+            /(?P<source_name>\w+)
+            /((?P<schema_name>\w+)-)?(?P<table_name>\w+)
+            (?:(?P<file_ext>.yaml|.sql|.manifest|/csv/(:?part-.*(:?\.gz)?|_SUCCESS)))$
+        """,
         re.VERBOSE,
     )
 
@@ -242,6 +242,8 @@ def _find_matching_files_from(iterable, pattern, return_success_file=False):
         match = file_names_re.search(filename)
         if match:
             values = match.groupdict()
+            if not values["schema_name"]:
+                values["schema_name"] = values["source_name"]
             target_table_name = TableName(values["source_name"], values["table_name"])
             if pattern.match(target_table_name):
                 file_ext = values["file_ext"]
--- a/python/etl/file_sets.py
+++ b/python/etl/file_sets.py
@@ -228,6 +228,42 @@ def _find_matching_files_from(iterable, pattern, return_success_file=False):
     file exists or that a SQL file is not present along with a manifest).
 
     Files ending in '_SUCCESS' or '_$folder$' are ignored (which are created by some Spark jobs).
+
+    >>> found = _find_matching_files_from([
+    ...     "/schemas/store/public-orders.yaml",
+    ...     "/data/store/public-orders.manifest",
+    ...     "/data/store/public-orders/csv/_SUCCESS",
+    ...     "/data/store/public-orders/csv/part-0.gz",
+    ...     "/schemas/dw/orders.sql",
+    ...     "/schemas/dw/orders.yaml",
+    ... ], pattern=TableSelector(["store.orders", "dw"]), return_success_file=True)
+    >>> files = dict(found)
+    >>> files["/schemas/store/public-orders.yaml"]["file_type"]
+    'yaml'
+    >>> files["/schemas/store/public-orders.yaml"]["source_name"]
+    'store'
+    >>> files["/schemas/store/public-orders.yaml"]["schema_name"]
+    'public'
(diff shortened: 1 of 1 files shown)


soumyadsanyal:
> I'm revisiting our filename patterns in the context of supporting file formats other than CSV. This made me wonder whether we still should keep that awkward duplication of the schema name for transformations. So instead of:
> 
> ```
> schemas/dw/dw-fact_users.sql
> schemas/dw/dw-fact_users.yaml
> ```
> 
> we'd just have
> 
> ```
> schemas/dw/fact_users.sql
> schemas/dw/fact_users.yaml
> ```
> 
> How to test -- first on `next`
> 
> * Change some files that are for transformations by dropping the second schema name. For the example above:
> 
> ```
> mv schemas/dw/dw-fact_order.sql schemas/dw/fact_order.sql
> mv schemas/dw/dw-fact_order.yaml schemas/dw/fact_order.yaml
> ```
> 
> * While on next, run the `ls` command _after building the image_:
> 
> ```
> arthur.py ls dw.fact_order
> ```
> 
> and watch the error pop up:
> 
> ```
> Bailing out: FileNotFoundError: Found no matching files in './schemas' for '['dw.fact_order']'
> ```
> 
> How to test -- then on this branch with the same change of file names from above:
> 
> * While on this branch, run the `ls` command _after building the image_:
> 
> ```
> arthur.py ls dw.fact_order
> ```
> 
> and observe the simple output
> 
> ```
> Schema: 'dw'
>     Table: 'dw.fact_order'
>         schemas/dw/fact_order.yaml
>         schemas/dw/fact_order.sql
> ```
> 
> For more testing, you should also be able to do:
> 
> ```
> arthur.py sync dw.fact_order
> arthur.py upgrade --dry-run dw.fact_order
> ```

❤️ -- specifically for 

- this PR description, 
- problem statement, 
- sample input / output flow and 
- before/after comparison to explain intended semantics.

python/etl/file_sets.py
```diff
@@ -231,17 +231,19 @@ def _find_matching_files_from(iterable, pattern, return_success_file=False):
     """
     file_names_re = re.compile(
         r"""(?:^schemas|/schemas|^data|/data)
-                                   /(?P<source_name>\w+)
-                                   /(?P<schema_name>\w+)-(?P<table_name>\w+)
-                                   (?:(?P<file_ext>.yaml|.sql|.manifest|/csv/(:?part-.*(:?\.gz)?|_SUCCESS)))$
-                               """,
+            /(?P<source_name>\w+)
+            /((?P<schema_name>\w+)-)?(?P<table_name>\w+)
```
soumyadsanyal, line 235: 👍 

Apply `?` operator to predicate.

python/etl/file_sets.py
```diff
@@ -231,17 +231,19 @@ def _find_matching_files_from(iterable, pattern, return_success_file=False):
...
+            (?:(?P<file_ext>.yaml|.sql|.manifest|/csv/(:?part-.*(:?\.gz)?|_SUCCESS)))$
+        """,
         re.VERBOSE,
     )
 
     for filename in iterable:
         match = file_names_re.search(filename)
         if match:
             values = match.groupdict()
+            if not values["schema_name"]:
```
soumyadsanyal, line 245: It looks like we're supporting the option to skip specifying the schema name, in which case we impute the source name to conform with the prior convention.

If so -- does it make sense to support the prior convention (source-name, schema-name, table-name) in the future? IIRC there are only a few schemas where the source name and schema name differ (e.g.: `yesterday`).

tvogels01: The source schema name matters for sources. So if you load something from an upstream database, say, `store` which has `orders` in the `public` schema, then the name is `store/public-orders` to show exactly which `orders` table in the `store` database we're referring to.

For transformations, this never made sense since they happen in the schema which is the name of the transformation schema.

That said, I'd like the file sets to be ignorant of source vs. transformations. So there's a simple fallback and that's all there is to it.

python/etl/file_sets.py
tvogels01: Good point, I'll add `?:` to not store this match.

soumyadsanyal: Oh I was mostly thinking out loud / parsing the change. 

Should have said "applies". And maybe also had some coffee?

python/etl/file_sets.py
soumyadsanyal: That makes sense, merci @tvogels01 !

PR #178, opened 2020-04-15
Skip non-data files when reporting data files

tvogels01:
Now that we've opened the spigot for data files and allow files other than "Sqoop"ed files (which start with `part` and end with `gz`), we run into the bug where a `_SUCCESS` or `manifest` file is accidentally picked up.  So we opened the spigot too far. This PR adds a better filter around what is a data file. (Random files might still be picked up... we'll deal with that when it happens.)


tvogels01:
We're woefully behind with tests :(

PR #179, opened 2020-04-22
Support extra IAM roles on the Redshift cluster

blohr-hs:
This adds some new parameters to the CloudFormation template for the Redshift cluster that allows the end user to assign up to 5 additional IAM roles to the cluster. The create, update, and delete paths were all tested with throw-away clusters.

The impetus of this is to allow us to capture manual role additions that have been made in CloudFormation parameters, formalizing them, as well as to let us add a new IAM role that we want to add soon.

### DIFF
--- a/cloudformation/dw_cluster.yaml
+++ b/cloudformation/dw_cluster.yaml
@@ -49,6 +49,31 @@ Parameters:
         Type: String
         Default: ""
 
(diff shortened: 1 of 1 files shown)


blohr-hs:
This has been tested in dev, and used in prod to update the cluster successfully.

PR #180, opened 2020-04-27
Add auto option for distribution style in H-D-W schemas

ynaim94:
## Description of changes

* For the schemas in H-D-W, it is now possible to set the distribution field to "auto", which is one of the distribution style permitted. This will tell Redshift  to automatically choose a distribution style for that table.

## Impacted areas

I identified two impacted areas:
  * Validation of the schemas against the table design.
  * Formatting the SQL query for the create table statement.

## Testing

* I ran a validation with arthur to check the scemas, skipping the dependencies and upstream check.
* I did not test the create table part.

## Risks

* The generation for create table statement was not tested.

### DIFF
--- a/python/etl/design/redshift.py
+++ b/python/etl/design/redshift.py
@@ -113,10 +113,8 @@ def build_table_attributes(table_design: dict) -> List[str]:
         if isinstance(distribution, list):
             ddl_attributes.append("DISTSTYLE KEY")
             ddl_attributes.append("DISTKEY ( {} )".format(join_column_list(distribution)))
-        elif distribution == "all":
-            ddl_attributes.append("DISTSTYLE ALL")
-        elif distribution == "even":
-            ddl_attributes.append("DISTSTYLE EVEN")
+        else:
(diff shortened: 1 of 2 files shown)


python/etl/design/redshift.py
```diff
@@ -117,6 +117,8 @@ def build_table_attributes(table_design: dict) -> List[str]:
             ddl_attributes.append("DISTSTYLE ALL")
         elif distribution == "even":
             ddl_attributes.append("DISTSTYLE EVEN")
+        elif distribution == "auto":
+            ddl_attributes.append("DISTSTYLE AUTO")
```
tvogels01, line 121: That'll work. What would you think of:
```
else:
    ddl_attributes.append("DISTSTYLE {}".format(distribution.upper())
```
instead of three branches?
We can trust that the validation makes sure that only valid values show up.

ynaim94: Yes I was thinking of that but hesitated because of readability, but I'll add that.

PR #181, opened 2020-05-07
Switch EMR nodes to spot instances instead of on-demand instances

jamessthompson:
Currently, EMR clusters in our pipelines utilize on-demand instances, by shifting to spot instances we can reduce node costs to below the reserved instance hourly rates while simultaneously avoiding their 24/7 requirement.

https://harrys.atlassian.net/browse/DOC-136
https://harrys.atlassian.net/browse/TECK-573

This change will break backwards compatibility with existing Arthur pipelines.

### DIFF
--- a/python/etl/config/settings.schema
+++ b/python/etl/config/settings.schema
@@ -119,6 +119,7 @@
                     "type": "integer",
                     "minimum": 1
                 },
+                "bid_price": { "type": "number" },
                 "managed_security_group": { "type": "string" }
             },
             "required": [ "instance_type", "managed_security_group" ],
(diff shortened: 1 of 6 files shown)


python/etl/config/settings.schema
```diff
@@ -119,6 +119,7 @@
                     "type": "integer",
                     "minimum": 1
                 },
+                "bid_price": { "type": "number" },
```
tvogels01, line 122: Please add a `"minimum"` -- might seem silly but there's no sanity check elsewhere.

tvogels01:
* The default settings are ignored when there are `resources` in a configuration file. I'll check whether to change Arthur to pull in defaults at a more granular level.
* Other than that, this appears good to go.

PR #187, opened 2020-05-28
Use epoch time (seconds) instead of decimal date

tvogels01:
This solidifies partitioning tables during the extract step when we want to partition by a timestamp or date. The current code unfortunately leads to skewed files since the partition key is computed as a numeric interpretation of the date (`20200529184819`) instead of seconds (1590778099).
A second update here makes sure to test on the column types in their generic form (`date`, `timestamp`) instead of their SQL types (same/similar for dates and timestamps but varying for int, long, float, etc.)

Changes required:
* For any table with a partition key that's a date or timestamp, make sure that the generic `type` info reflects that it's a date or timestamp. A `string` value will no longer be supported as the partition key. This change can be made ahead of deploying this version.

### DIFF
--- a/python/etl/extract/sqoop.py
+++ b/python/etl/extract/sqoop.py
@@ -181,8 +181,8 @@ class SqoopExtractor(DatabaseExtractor):
         """
         if partition_key:
             column = fy.first(fy.where(relation.table_design["columns"], name=partition_key))
-            if column["sql_type"] in ("timestamp", "timestamp without time zone"):
-                quoted_key_arg = """CAST(TO_CHAR("{}", 'YYYYMMDDHH24MISS') AS BIGINT)""".format(partition_key)
+            if column["type"] in ("date", "timestamp"):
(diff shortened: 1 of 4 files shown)


python/etl/extract/sqoop.py
```diff
@@ -181,8 +181,8 @@ def build_sqoop_partition_options(
         """
         if partition_key:
             column = fy.first(fy.where(relation.table_design["columns"], name=partition_key))
-            if column["sql_type"] in ("timestamp", "timestamp without time zone"):
-                quoted_key_arg = """CAST(TO_CHAR("{}", 'YYYYMMDDHH24MISS') AS BIGINT)""".format(partition_key)
+            if column["type"] in ("date", "timestamp"):
+                quoted_key_arg = """CAST(DATE_PART('epoch', "{}") AS BIGINT)""".format(partition_key)
```
blohr-hs, line 185: Did you happen to test this out on Postgres 9.6 to make sure this works?

tvogels01: Yes. I ran the extract with this code on a database with 9.6.11.

PR #190, opened 2020-06-01
Use Spot instances for EMR cluster (take 2)

tvogels01:
We reverted using spot instances for the EMR cluster since a necessary role not automatically created. This is the second attempt to get this feature out. The role for the Data Pipeline and EMR cluster will now create the service-linked role.

See:
* Switch EMR nodes to spot instances instead of on-demand instances #181
* Hotfix revert spot instances #189


tvogels01:
Searching for the service role after deploying the change:
```
Path
/aws-service-role/spot.amazonaws.com/
Creation time
2020-06-12 13:55 EDT
```
Which is to say the role got created on-demand.

PR #193, opened 2020-06-10
Add -y (assume yes) to upload script

tvogels01:
When using the Arthur deploy script, which uploads the ETL distribution file to S3, inside another script, it's helpful to avoid the prompt from the deploy script. So this PR adds a `-y` option (similar to yum's "--assume-yes"). Note that this is now the default when deploying Arthur ETL using the Docker image with `bin/deploy_arthur.sh`.

### DIFF
--- a/bin/docker_run.sh
+++ b/bin/docker_run.sh
@@ -142,21 +142,25 @@ case "$action" in
             --env ARTHUR_DEFAULT_PREFIX="$target_env" \
             $profile_arg \
             "arthur:$tag" \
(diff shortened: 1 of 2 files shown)


tvogels01:
@ynaim94 The change is on `upload.sh`. So you'll have to run that. The `deploy_arthur.sh` script leverages the new command line option from `upload.sh` but itself doesn't have that option.

It'll be easiest for us to merge this PR first and then you can easily test `upload.sh` from within the Docker container: Fix search for distribution (setup.py) for Docker image #192

PR #197, opened 2020-06-14
Help with pgpass file

tvogels01:
The use of `create_user` or `add_user` commands requires a `~/.pgpass` file (or whatever `$PGPASSFILE` points to). This can create some confusion and extra steps. So this PR makes first the error messages more information (and log lines also provide next steps)., then adds a default `.pgpass` file to show the expected format. Now the only exception message should be about a missing line, and that message is preceded by a line which can by added to `.pgpass` (along with a suitable and secure password!).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,10 @@ echo \n\
 echo "Environment settings:"\n\
 arthur.py settings object_store.s3.* version' > /root/.bashrc
 
+# Create an empty .pgpass file to help with create_user and update_user commands.
+RUN echo '# Format to set password (used by create_user and update_user): *:5439:*:<user>:<password>' > /root/.pgpass \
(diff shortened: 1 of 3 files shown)


soumyadsanyal:
@tvogels01 

I looked at the spec for `pgpass` : https://www.postgresql.org/docs/9.3/libpq-pgpass.html

and noticed that we wouldn't really have room to add metadata to construct groups or add users to those groups. Please disregard the q for now!

tvogels01:
The intention of `~/.pgpass` is set the password in a repeatable way when running `create_user`.

The construction of groups should go into the settings file.
See https://github.com/harrystech/arthur-redshift-etl/blob/v1.23.1/python/etl/config/settings.schema#L167

PR #134, opened 2019-07-17
Expect the unexpected S3Query Exception

tvogels01:
We received a new kind of error from Redshift running into troubles:
```
psycopg2.errors.InternalError_: S3Query Exception Connection failed, retries exceeded.
DETAIL:  
  -----------------------------------------------
  error:  S3Query Exception Connection failed, retries exceeded.
  code:      15005
  context:   
  query:     15067689
  location:  dory_util.cpp:844
  process:   query1_555_15067689 [pid=42951]
  -----------------------------------------------
```

This PR updates the strings we're looking for to detect spectrum errors with problems connecting to S3.

PR #198, opened 2020-06-28
DOC-210 Spectrum retries via pgcode list

### LINKED ISSUE #164: Test the error code, not error string
# Summary

See https://github.com/harrystech/arthur-redshift-etl/pull/134 -- we should more generally test the pg error code that's returned instead of the error string.

# Details

(need examples)

marwamc-hs:
For [Spectrum scan errors](https://docs.aws.amazon.com/redshift/latest/dg/c-spectrum-troubleshooting.html#spectrum-troubleshooting-retries-exceeded) caused by too many small files, we setup Arthur to retry if the PGCODE is a known value.

Previously, we have used the the PGERROR name, which has proved to be unreliable.

NOTE: The current Spectrum Scan Error `PGCODE=15001` is not even listed in the official [AWS List of Known Load Errors](https://docs.aws.amazon.com/redshift/latest/dg/r_Load_Error_Reference.html). We need to identify more Spectrum/S3 timeout PGCODE errors and add them to the list.

Testing inside Docker container:
```
$ python3
>>> etl.config.load_config(['... add path to config file ...'])
>>> etl.config.get_config_list('arthur_settings.retriable_error_codes')
```

Closes #164

### DIFF
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -78,11 +78,11 @@ def get_config_int(name: str, default: Optional[int] = None) -> int:
         return int(value)
 
 
-def get_config_list(name: str) -> List:
+def get_config_list(name: str) -> List[int]:
     """
-        Lookup a configuration value that is a List.
+    Lookup a configuration value that is a List.
     """
-    return _mapped_config.get(name)
+    return list(map(int, get_config_value(name).split(",")))
 
 
 def set_config_value(name: str, value: str) -> None:
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -82,7 +82,7 @@ def get_config_list(name: str) -> List:
     """
         Lookup a configuration value that is a List.
     """
-    return []
+    return _mapped_config.get(name)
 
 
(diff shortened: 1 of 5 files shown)


tvogels01:
Given the limitations of the current "settings" system in Arthur (support for maps of `str` to `str` only), I could see us taking a 3 step approach.

1. (sad but...) Use a string to represent the list of codes, e.g. `8001,15001`. We then use `list(map(int, codes.split(',')))` when returning a list from config. There are clear danger signs and we'd miss out on validation but we'd get this code into production to reduce interruptions / failed ETLs.
2. (slightly better) We add the option in settings to support a map from `str` to `Union[str,List[str]]`. So we'd end up with `[ '8001', '15001' ]. We can start validating now against having a list. We'd have to update `show_vars` and friends.
3. (even better) Allow `Union[str,int,List[str],List[int]]`. :-)
4. profit

Note that we seem to also need 8001 as an error code. But let's check our stl load errors again.

tvogels01:
List of previously encountered error messages: Expect the unexpected S3Query Exception #134

PR #200, opened 2020-07-15
Stop using spot instances for rebuild

tvogels01:
We've had a couple issues with rebuilds stopping because the EMR cluster loses its spot instances before extraction is done.

This PR (temporarily) reverts using spot instances for rebuilds until we can switch to using a spot fleet.


tvogels01:
Yes. omitting the bid price means on-demand.

See https://github.com/harrystech/arthur-redshift-etl/pull/190/files

PR #206, opened 2020-07-20
Support falling back to AWS_DEFAULT_PROFILE

tvogels01:
This allows to set `AWS_DEFAULT_PROFILE` as an alternative to `AWS_PROFILE` before starting the Docker container.
The idea here is that `AWS_DEFAULT_PROFILE` may be set in the shell startup, _e.g._ `~/.zshrc`, along with
`ARTHUR_DEFAULT_PREFIX` and `DATA_WAREHOUSE_CONFIG` to simplify running `bin/run_arthur.sh` while
still having the option to easily set a new AWS profile using either `AWS_PROFILE` or the `-p` command line option.

### DIFF
--- a/bin/docker_run.sh
+++ b/bin/docker_run.sh
@@ -36,8 +36,11 @@ configured to use <config_dir>.
 
 The <config_dir> defaults to \$DATA_WAREHOUSE_CONFIG.
 The <target_env> defaults to \$ARTHUR_DEFAULT_PREFIX (or \$USER if not set).
+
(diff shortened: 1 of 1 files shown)


tvogels01:
Leaving a breadcrumb of our discussion here. The assumption is that if somebody explicitly sets `AWS_PROFILE` to empty, then that should be respected in the evaluation of the value of `profile`.  So `unset AWS_PROFILE` leads to `AWS_DEFAULT_PROFILE` being evaluated, but `AWS_PROFILE=` does not.

PR #208, opened 2020-07-26
Move .bashrc into etc/

tvogels01:
The motivation here is to have the command line prompt to show the "prefix" (which is usually the user's name or the environment, such as `tom` or `production`) -- reduces second guessing whether you're in the right environment.

This PR also adds an alias `develop` to make developing inside the Docker container easier when one is working on scripts or the like and a `python setup.py develop` in the correct directory is required.

Since there continues to be stuff added to the `.bashrc` file, this is now broken out of the `Dockerfile` into the `etc` directory.

Example:
```
bin/run_arthur.sh
...
(venv) (aws:data-dev, prefix=tom) # 
```

### DIFF
--- a/etc/default.bashrc
+++ b/etc/default.bashrc
@@ -1,0 +1,17 @@
+# This is the default .bashrc file inside a container.
+
+PS1='(aws:$AWS_PROFILE, prefix=$ARTHUR_DEFAULT_PREFIX) \$ '
+
+source /tmp/redshift_etl/venv/bin/activate
+PATH=$PATH:/tmp/redshift_etl/bin
+
+# Useful when developing Arthur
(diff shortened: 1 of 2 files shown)


etc/default.bashrc
```diff
@@ -0,0 +1,17 @@
+# This is the default .bashrc file inside a container.
+
+PS1='(aws:$AWS_PROFILE, prefix=$ARTHUR_DEFAULT_PREFIX) \$ '
```
marwamc-hs, line 3: add current arthur version...
too much? 😄

tvogels01: I don't know an easy way for that. I don't want to run `arthut.py --version` every time.

PR #210, opened 2020-08-11
Add automatic linting as an Action

tvogels01:
Playing around with GitHub Actions --- this will give us linting for PRs.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -30,7 +30,7 @@ jobs:
         black --config ./etc/pyproject.toml --check python/
     - name: Lint with isort
       run: |-
-        isort --settings-path ./etc --check-only --recursive python/
+        isort --settings-path ./etc --check-only python/
     - name: Lint with pycodestyle
       run: |-
         pycodestyle python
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -1,3 +1,8 @@
+[mypy]
+python_version = 3.5
+ignore_missing_imports = True
+strict_optional = True
+
 [pycodestyle]
 # Seetings are compatible with black, see https://github.com/psf/black/issues/354
```
marwamc-hs, line 7: `seetings` --> `settings`

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,41 @@
...
+        black --config ./etc/pyproject.toml --check python/
+    - name: Lint with isort
+      run: |-
+        isort --settings-path ./etc --check-only python/
+    - name: Lint with pycodestyle
+      run: |-
+        pycodestyle python
+    - name: Run type checker
+      run: |-
+        mypy python
```
marwamc-hs, line 41: since no dependencies installed, this only lints arthur code.

tvogels01: You mean compared to d/e/m (data eng microservices) setup? Yes, requirements are not installed inside the `python` directory so this doesn't accidentally pick up other code.

PR #212, opened 2020-08-12
Add running doc tests as an Action

tvogels01:
Follow-up and addition to PR #210 -- this adds running the doctests when changes are pushed.

### DIFF
--- a/python/etl/selftest.py
+++ b/python/etl/selftest.py
@@ -123,4 +123,6 @@ def run_tests() -> None:
 
 
 if __name__ == "__main__":
-    run_tests()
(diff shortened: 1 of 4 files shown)


python/etl/selftest.py
```diff
@@ -123,4 +123,6 @@ def run_tests() -> None:
 
 
 if __name__ == "__main__":
-    run_tests()
+    # Running "python3 -m python.etl.selftest" will only run doc tests.
+    # Use "run_tests.py" to run all of the tests.
```
marwamc-hs, line 127: I don't see a `run_test.py` script...

tvogels01: Look for it in the `setup.py` file!

PR #213, opened 2020-08-12
Update VPC and CopyRole to allow unloading to any bucket

soumyadsanyal:
- Like it says on the tin.

Why: we're unloading to a new bucket in development. We don't want to catalog all the allowed buckets.

Consequences: The cluster (using CopyRole) will be able to unload warehouse data to any bucket in S3, including public buckets that permits writes.

### DIFF
--- a/cloudformation/dw_vpc.yaml
+++ b/cloudformation/dw_vpc.yaml
@@ -240,7 +240,7 @@ Resources:
                             "s3:PutObject*"
                         ],
                         "Resource": [
-                            !Sub "arn:aws:s3:${AWS::Region}:${AWS::AccountId}:*"
+                            !Sub "arn:aws:s3::${AWS::AccountId}:*"
                         ]
                     }
                 ]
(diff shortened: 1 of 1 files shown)


cloudformation/dw_vpc.yaml
```diff
@@ -838,9 +834,7 @@ Resources:
                                 - "s3:DeleteObject"
                                 - "s3:PutObject"
                             Resource:
-                                - !Sub "arn:aws:s3:::${ObjectStore}/*"
-                                - !Sub "arn:aws:s3:::${DataLake}/*"
-                                - !If [ "ValidNotEmptySecondaryDataLake",  !Sub "arn:aws:s3:::${SecondaryDataLake}/*",  !Ref "AWS::NoValue" ]
+                                - !Sub "arn:aws:s3:::*"
```
tvogels01, line 837: I'd like to add something like this:
```
        "Condition": {"StringLike": {"aws:Tags/user:project": [
            "warehouse"
        ]}}
```

cloudformation/dw_vpc.yaml
```diff
@@ -232,17 +232,15 @@ Resources:
...
                         "Principal": "*",
                         "Action": [
                             "s3:DeleteObject*",
                             "s3:PutObject*"
                         ],
                         "Resource": [
-                            !Sub "arn:aws:s3:::${ObjectStore}/*" ,
-                            !Sub "arn:aws:s3:::${DataLake}/*" ,
-                            !If [ "ValidNotEmptySecondaryDataLake",  !Sub "arn:aws:s3:::${SecondaryDataLake}/*", !Ref "AWS::NoValue" ]
+                            !Sub "arn:aws:s3:::*"
```
tvogels01, line 243: We don't need the `!Sub` any more ...

cloudformation/dw_vpc.yaml
tvogels01: The additional condition will be part of the data lake vs. object store refactor.

PR #214, opened 2020-08-14
Rethink docker

tvogels01:


Dockerfile
```diff
@@ -21,42 +22,47 @@ RUN yum install -y \
...
-# From here, bind-mount your data warehouse code directory to /data-warehouse.
-# All of the normal Arthur configuration for accessing and managing the data
-# warehouse is assumed to have already been set up in that directory.
+# The data warehouse (with schemas, config, etc.) will be mounted here:
+WORKDIR /opt/data-warehouse
 
-ENTRYPOINT ["/arthur-redshift-etl/bin/entrypoint.sh"]
-CMD ["/bin/bash"]
+ENTRYPOINT ["/opt/src/arthur-redshift-etl/bin/entrypoint.sh"]
+CMD ["/bin/bash", "--login"]
```
soumyadsanyal, line 68: 👍 

> When bash is invoked as an interactive login shell, or as a non-interactive shell with the --login option, it first reads and executes commands from the file /etc/profile, if that file exists. After reading that file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and executes commands from the first one that exists and is readable. The --noprofile option may be used when the shell is started to inhibit this behavior.

PR #216, opened 2020-08-15
Add flake8 and update pre-commit hook

tvogels01:
This PR replaces `pycodestyle` with `flake8` as the recommended linter since `flake8` has more options to be extended, in particular, `flake8` adds checks on doc strings and complexity.
* As a bonus, we'll now check linters and type checker in parallel -- the configuration for `mypy` and thus its runtime will get more complex, e.g by adding `mypy-boto3`.
* The file `setup.py` is now also being linted.
* There are small changes to some code to avoid having to ignore `D211` from the start. Other error codes are ignored for now to allow bringing in the tool and then we can cleanup over time.
* Note that the "self test" still runs a PEP8 check using `pycodestyle`. That's left for a future refactor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,14 @@ end_of_line = lf
 insert_final_newline = true
 max_line_length = 100
 
-[*.py]
+[*.md]
+indent_size = 4
+indent_style = space
+max_line_length = 80
+tab_width = 4
+trim_trailing_whitespace = true
+
+[*.{py,sh}]
 indent_size = 4
 indent_style = space
 tab_width = 4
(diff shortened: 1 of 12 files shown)


.editorconfig
```diff
@@ -6,7 +6,14 @@ end_of_line = lf
 insert_final_newline = true
 max_line_length = 100
 
-[*.py]
+[*.md]
+indent_size = 4
```
soumyadsanyal, line 10: 2 or 4?

tvogels01: When I started this project, I picked 4 because my eyes are used to 4 for Python.
Then people using Ruby (2 spaces) and Scala (2 spaces) chimed in and pushed other projects to have 2 spaces in YAML files. 
The two arthur repos are the holdouts...

PR #218, opened 2020-08-15
Update issue templates

tvogels01:
This is a test of splitting our issues by bugs and features. (We don't receive a lot of either so that's more a test of the GitHub feature that we might learn from.)


marwamc-hs (approved):
LGTM

PR #263, opened 2020-10-29
Change PercentageDiskSpaceUsed threshold from 65% to 85%

tvogels01:
This sets the threshold for the low-space alert in Redshift to 85%. (Next step: make this a parameter.)

This alert is a high-water line that may trigger during the ETL or normal use of the warehouse. Leaving just 15% of margin is expected to be enough to start checking queries or adding a note before the ETL fails when running out of disk space. (If we trigger too early, we'd buy an additional cluster node too early.)

Note that the ETL builds staging tables that are store in addition to the "working set" and any backup tables so the disk use from tables accessed by normal users is much less than 85% and more like 25% to 33%.

(We'll have to add an alert for low-water line so that we know when the disk space used up by standard position and backups reaches 60%.)

Deployment (e.g. for `prod`):
```
cloudformation/do_cloudformation.sh create-change-set dw_cluster prod \
VpcStackName=UsePreviousValue \
    MasterUsername=UsePreviousValue \
    MasterUserPassword=UsePreviousValue \
    NodeType=UsePreviousValue \
    NumberOfNodes=UsePreviousValue \
    QueryConcurrency=UsePreviousValue \
    SnapshotIdentifier=UsePreviousValue \
    AdditionalClusterIAMRole1=UsePreviousValue \
    AdditionalClusterIAMRole2=UsePreviousValue \
    AdditionalClusterIAMRole3=UsePreviousValue \
    AdditionalClusterIAMRole4=UsePreviousValue \
    AdditionalClusterIAMRole5=UsePreviousValue \
    PreferredMaintenanceWindow=UsePreviousValue

aws cloudformation describe-change-set --change-set-name "add Id from create cs command here"
aws cloudformation execute-change-set --change-set-name "add Id from create cs command here"
```

### DIFF
--- a/cloudformation/dw_cluster.yaml
+++ b/cloudformation/dw_cluster.yaml
@@ -203,7 +203,7 @@ Resources:
             Namespace: AWS/Redshift
             Statistic: Average
             ComparisonOperator: GreaterThanThreshold
-            Threshold: 65
+            Threshold: 75
             Period: 300
             EvaluationPeriods: 3
             Dimensions:
--- a/cloudformation/dw_cluster.yaml
+++ b/cloudformation/dw_cluster.yaml
@@ -203,7 +203,7 @@ Resources:
             Namespace: AWS/Redshift
             Statistic: Average
             ComparisonOperator: GreaterThanThreshold
-            Threshold: 75
+            Threshold: 85
             Period: 300
             EvaluationPeriods: 3
             Dimensions:


soumyadsanyal (approved):
👍

soumyadsanyal:
Related to, but does not directly follow from https://harrys.atlassian.net/browse/DENG-416.

@tvogels01 Could you please document the reasoning behind this change?

marwamc-hs (approved):
LGTM
We'll get to that bridge...

PR #304, opened 2021-01-15
DENG-528: Upgrade Python Arthur

willcampbell433:
**Description**
This PR addresses the following:

1. Updates the AMI used for the EMR cluster to 6.1.0 which runs Python 3.7.9 – the same Python version used for the EC2 and local Docker image
2. Modifies the local Dockerfile to use the matching Amazon Linux 2 image
3. Changes `pip install` commands in the Dockerfile to use `python3 -m pip install` instead

**JIRA ticket**

https://harrys.atlassian.net/browse/DENG-528

### DIFF
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -44,7 +44,7 @@
             "region": "us-east-1"
         },
         "EC2": {
-            "image_id": "ami-0623133d3181bb2d9",
+            "image_id": "ami-0be2609ba883822ec",
             "instance_type": "t2.small"
         },
         "EMR": {
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -40,11 +40,11 @@
             "region": "us-east-1"
         },
         "EC2": {
-            "image_id": "ami-a4c7edb2",
+            "image_id": "ami-01d7070812ee4de14",
             "instance_type": "t2.small"
         },
         "EMR": {
-            "release_label": "emr-5.3.0",
+            "release_label": "emr-6.2.0",
             "master": {
                 "bid_price": 10.0,
                 "instance_type": "m4.2xlarge",
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -40,11 +40,11 @@
             "region": "us-east-1"
         },
         "EC2": {
-            "image_id": "ami-01d7070812ee4de14",
+            "image_id": "ami-0623133d3181bb2d9",
             "instance_type": "t2.small"
         },
         "EMR": {
-            "release_label": "emr-6.2.0",
+            "release_label": "emr-6.1.0",
             "master": {
                 "bid_price": 10.0,
                 "instance_type": "m4.2xlarge",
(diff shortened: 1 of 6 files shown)


willcampbell433:
Changes look good 👍 

Testing acceptance criteria:

- [ ] Successful validation pipeline: `install_validation_pipeline.sh`
- [ ] Successful extraction pipeline: `install_extraction_pipeline.sh tallboy.redshift_pipe_cleaner`
- [ ] Successful upgrade pipeline: `install_upgrade_pipeline.sh tallboy.redshift_pipe_cleaner`
- [ ] Build local Docker image and check `watch` and `psql` packages
- [ ] Launch EC2 instance and check packages: `launch_ec2_instance.sh`
- [ ] Launch EMR cluster and check packages: `launch_emr_cluster.sh`

willcampbell433:
LGTM. Marking as "ready for review". Let's test in dev tonight. @tvogels01

python/etl/config/default_settings.yaml
```diff
@@ -40,11 +40,11 @@
             "region": "us-east-1"
         },
         "EC2": {
-            "image_id": "ami-a4c7edb2",
+            "image_id": "ami-0623133d3181bb2d9",
```
ynaim94, line 43: This seems to be a private image in the dev account, wouldn't that be a problem to deploy in prod? Or are we going to make it available in prod after testing?

python/etl/config/default_settings.yaml
```diff
@@ -40,11 +40,11 @@
...
             "instance_type": "t2.small"
         },
         "EMR": {
-            "release_label": "emr-5.3.0",
+            "release_label": "emr-6.1.0",
```
ynaim94, line 47: Why not the latest release (6.2)?

tvogels01: When trying to use 6.2, the extraction failed. Looks like two tools installed their version of "open csv" and then the bootstrap step fails with:
```
+ sudo ln -s /usr/lib/hive/lib/opencsv-2.3.jar /usr/lib/hive/lib/opencsv-3.9.jar /mnt/taskRunner/open-csv.jar
```

python/etl/config/default_settings.yaml
tvogels01: Bizarre. This is the AMI ID that came up when I started an EMR cluster by hand for 6.1

tvogels01: @willcampbell433 Back to you :)

So I picked the AMI from the instance that was running in the EMR cluster. But Youssef found that this is marked as "private" in the AWS AMI images. Let's not use that then.

The recipe then is to find an AMI image and Docker tag that match each other.
* Please find the AMI image most similar to `2.0.20201218.1`
* And then test that against an extraction, upgrade and validation pipeline.

We don't need to be too paranoid about matching AMI versions just such that when we've developed a new version in Docker, it should reliably run the same way (based on the same Python version and package versions) in EMR and on the EC2 box.

willcampbell433: updated to use the public image from AWS

python/etl/templates/extraction_pipeline.json
```diff
@@ -60,7 +60,7 @@
                 "${resources.VPC.whitelist_security_group}"
             ],
             "bootstrapAction": "s3://${object_store.s3.bucket_name}/${object_store.s3.prefix}/bin/bootstrap.sh,${object_store.s3.bucket_name},${object_store.s3.prefix}",
-            "applications": [ "Spark", "Ganglia", "Zeppelin", "Sqoop" ],
+            "applications": [ "Ganglia", "Sqoop", "Zeppelin" ],
```
tvogels01, line 63: Note to self ... we need to get rid of the other Spark vestiges.

tvogels01 (approved):
LGTM

PR #310, opened 2021-01-20
Improve create user command

tvogels01:
This PR makes creating and updating a users a bit easier.
* There is a dedicated command to create a valid password.
* It's now OK when updating a user for that user to not be in the `~/.pgpass` file.

Example steps:
```
REDSHIFT_USER=arthur
REDSHIFT_PASSWORD=$( arthur.py show_random_password )

echo "*:5439:*:$REDSHIFT_USER:$REDSHIFT_PASSWORD" | tee -a ~/.pgpass

arthur.py create_user $REDSHIFT_USER

# The next command succeeds even if you remove the corresponding line from ~/.pgpass
arthur.py update_user --add-user-schema $REDSHIFT_USER
```

Notes
* Also moves the default for the schemas dir into a more appropriate place.
* Updates log output to show more info.

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -1265,7 +1265,7 @@ class RenderTemplateCommand(SubCommand):
         parser.set_defaults(log_level="CRITICAL")
         add_standard_arguments(parser, ["prefix"])
         group = parser.add_mutually_exclusive_group(required=True)
-        group.add_argCRITICALument("-l", "--list", help="list available templates", action="store_true")
+        group.add_argument("-l", "--list", help="list available templates", action="store_true")
         group.add_argument("template", help="name of template", nargs="?")
         parser.add_argument("-t", "--compact", help="produce compact output", action="store_true")
 
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -223,6 +223,7 @@ class FancyArgumentParser(argparse.ArgumentParser):
 
 
 def isoformat_datetime_string(argument):
+    # "isoformat" is used as a verb here.
     return datetime.strptime(argument, "%Y-%m-%dT%H:%M:%S")
 
 
@@ -282,6 +283,7 @@ def build_full_parser(prog_name):
     for klass in [
         # Commands to deal with data warehouse as admin:
         InitializeSetupCommand,
+        ShowRandomPassword,
         CreateUserCommand,
         UpdateUserCommand,
         # Commands to help with table designs and uploading them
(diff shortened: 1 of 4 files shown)


python/etl/commands.py
```diff
@@ -541,25 +543,42 @@ def callback(self, args, config):
...
+class ShowRandomPassword(SubCommand):
+    def __init__(self):
+        super().__init__(
+            "show_random_password",
+            "show a random password compatible with Redshift",
+            "Show a random password with upper-case, lower-case and a number.",
+        )
+
+    def add_arguments(self, parser):
+        parser.set_defaults(log_level="CRITICAL")
```
ynaim94, line 555: I am not sure if this is the pattern you're following when you moved `parser.set_defaults(table_design_dir="./schemas")` from `add_arguments()` to `add_to_parser()`, but should this line go in add_to_parser() as well? Like you did here:
```
class MonitoredSubCommand(SubCommand):
    """
    A sub-command that will also use monitors to update some event table
    """

    def add_to_parser(self, parent_parser) -> argparse.ArgumentParser:
        parser = super().add_to_parser(parent_parser)
        parser.set_defaults(use_monitor=True)
        return parser
```

python/etl/commands.py
```diff
@@ -570,25 +589,22 @@ def callback(self, args, config):
...
 class UpdateUserCommand(SubCommand):
     def __init__(self):
-        random_password = uuid.uuid4().hex
-        example_password = random_password[:16].upper() + random_password[16:].lower()
         super().__init__(
             "update_user",
             "update user's group, password, and path",
             "Update an existing user with group membership, password, and search path."
-            " Note that you have to have set a password for the user in your .pgpass file"
+            " Note that you have to set a password for the user in your '~/.pgpass' file"
```
ynaim94, line 596: Is that still the case from your PR description?

python/etl/db.py
```diff
@@ -371,21 +372,29 @@ def _get_encrypted_password(cx, user):
...
 
-def alter_password(cx, user):
+def alter_password(cx, user, ignore_missing_password=False):
     password = _get_encrypted_password(cx, user)
+    if password is None:
+        logger.warning("Failed to find password in PGPASSFILE for '%s'", user)
+        if not ignore_missing_password:
+            raise ETLRuntimeError("password missing from PGPASSFILE for user '{}'".format(user))
+        return
     execute(cx, """ALTER USER "{}" PASSWORD %s""".format(user), (password,))
```
ynaim94, line 398: Unrelated to the PR but how does the user know what is his new password?

python/etl/commands.py
tvogels01: Yes/no :-) The line must be present in case you want to update the password. I'll fix the description.

python/etl/commands.py
tvogels01: Another command for comparison is: `ShowHelpCommand`

The `MonitoredSubCommand` isn't a command that you'd run from the command line. It's just a parent for some commands.

I moved the `parser.set_defaults(table_design_dir="./schemas")` into the `add_to_parser` from `add_standard_arguments` because the former is invoked for all commands while the second isn't.

python/etl/db.py
tvogels01: I'm hoping the person running this command will share the password with the user.

Overall, we need to add the option of storing the password in some secure storage (AWS parameter store?) since the changes to the `~/.pgpass` file are lost when you exit the Docker container.

ynaim94 (approved):
LGTM

PR #311, opened 2021-01-20
Arthur can parse passwords containing the `=` sign.

soumyadsanyal:
We tripped over a password containing an `=` sign. The regex currently used by Arthur seemed to not allow that. Turns out that we need to URL-encode special characters in the DSN. Per documentation, e.g. of Postgres

> Percent-encoding may be used to include symbols with special meaning in any of the URI parts, e.g., replace = with %3D.

(We're making the choice here to not support escaping using `{ ... }`.)

This PR changes the regex that splits the DSN to expect the URI to be in ASCII and percent-encoded when using special characters.

Example connection string that now works correctly:
```
postgres://user:secret%3Dspice@db.example.com:5432/app?sslmode=require
```
(where the password is `secrect=spice`).

### DIFF
--- a/python/etl/db.py
+++ b/python/etl/db.py
@@ -51,7 +51,7 @@ def parse_connection_string(dsn: str) -> Dict[str, str]:
     # Now they have two problems.
     dsn_re = re.compile(
         r"""(?:jdbc:)?(?P<subprotocol>redshift|postgresql|postgres)://  # accept either type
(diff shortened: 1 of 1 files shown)


tvogels01 (commented):
Oops. Forgot to approve before merge.

soumyadsanyal:
Thank you @tvogels01 ! 🙏

PR #313, opened 2021-01-24
Pin pip and python versions (also in Docker image)

tvogels01:
This generally sets the Python version to 3.7 and PIP version to 20.3.4. The min Python version that's expected now is 3.6. (We're likely to start using `f`-strings!)

It seems that we no longer need `gcc` so that's removed from the installed packages.

Also, small cleanup in `Dockerfile` to use `PROJ_NAME` just like `boostrap.sh` does.


Dockerfile
```diff
@@ -7,9 +7,11 @@
 
 FROM amazonlinux:2.0.20201218.1
 
+# See the same settingin bin/bootstrap.sh
```
willcampbell433, line 10: small typo... should be "setting in"

willcampbell433 (approved):
LGTM

PR #314, opened 2021-01-24
Docstring changes for future max line width

tvogels01:
After these changes, we should have only E501s (line too long) left when re-formatting for 100 line width. (We'll be moving towards tighter formatting in arthur etl and arthur tools.)

Check using:
```
flake8 --config etc/future_setup.cfg --extend-ignore E501 python setup.py
```

Automatic reformatting will happen using:
```
black --config ./etc/future_pyproject.toml python setup.py
```


willcampbell433:
Just comment and linting updates so LGTM 👍🏻

willcampbell433 (approved):

PR #325, opened 2021-01-31
Improve tags of EC2 instances, reduce launch noise

tvogels01:
Motivation: use `jq` to reduce the output from the launch script to meaningful values:
```
    jq '.Instances[] | {Monitoring: .Monitoring, InstanceId: .InstanceId, ImageId: .ImageId}'
```

Related changes:
* The bootstrap script now only change the `Name` tag on instances for those from an EMR cluster.
* One-off instances set the name in the launch script and instances in a data pipeline get their `Name` reset by the data pipeline system.
* This change popped into this PR because it was confusing why the launch script didn't have control over `Name`.

Follow-up:
* We should add more tags to the data pipeline (like environment or bucket), knowing that we can't do this on the instance level.

General changes
* Change formatting to follow our `.editorconfig`
* Use descriptive tokens in "here" docs
* Prefer `$( ... )` over backticks
* Use `function_name() {`

### DIFF
--- a/python/etl/templates/ec2_instance.json
+++ b/python/etl/templates/ec2_instance.json
@@ -21,7 +21,7 @@
             "Tags": [
                 {
                     "Key": "Name",
-                    "Value": "redshift_etl (running bootstrap...)"
+                    "Value": "Arthur ETL (Initializing...)"
                 },
                 {
                     "Key": "user:project",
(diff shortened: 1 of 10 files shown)


python/etl/templates/ec2_instance.json
```diff
@@ -21,7 +21,7 @@
             "Tags": [
                 {
                     "Key": "Name",
-                    "Value": "redshift_etl (running bootstrap...)"
+                    "Value": "Arthur ETL (Initializing...)"
```
willcampbell433, line 24: I definitely like this over the old name :)

willcampbell433 (approved):
LGTM

PR #326, opened 2021-01-31
Escape . from schema.table as schema[.]table

tvogels01:
We're rewriting queries while building them so that we can build them in a staging schema. A bug crept in when a CTE had the same name as an upstream table on which the target table depended on.

Example:
```
WITH dw_fact AS (
  SELECT ... 
  FROM dw.fact
)
SELECT ..
FROM dw_fact
```

Before this fix:
```
WITH etl_staging$dw.fact AS (
  SELECT ... 
  FROM etl_staging$dw.fact
)
SELECT ..
FROM etl_staging$dw.fact
```

After this fix:
```
WITH dw_fact AS (
  SELECT ... 
  FROM etl_staging$dw.fact
)
SELECT ..
FROM dw_fact
```

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -210,10 +210,12 @@ class LoadableRelation:
     def query_stmt(self) -> str:
         stmt = self._relation_description.query_stmt
         if self.use_staging:
-            # Rewrite the query to use staging schemas:
+            # Rewrite the query to use staging schemas by changing identifiers to their staging
+            # version. This requires all tables to be fully qualified. There is a small chance
+            # that we're too aggressive and change a table name inside a string.
             for dependency in self.dependencies:
                 staging_dependency = dependency.as_staging_table_name()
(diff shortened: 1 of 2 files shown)


ynaim94 (commented):
Code looks good but how to test it?

tvogels01:
The test should follow the description of the bug: pick a transformation with a CTE and rename the CTE to be similar to a table that the transformation depends on. So if you start from the transformation `dw.fact_order` which depends on `dw._dim_user`, then make sure that the `dw.fact_order` transformation has a CTE called `dw_dim_user` (which could be just `SELECT 1`). Before this fix, this will get renamed to `dw$dim_user` but with this fix, its name will remain as `dw_dim_user`.
```
arthur.py sync dw.fact_order 
arthur.py upgrade dw.fact_order --with-staging --dry-run --only-selected
```
Check the `arthur.log` file to confirm.

Before:
```
WITH etl_staging$dw.dim_user AS (SELECT 1)
```
After:
```
WITH dw_dim_user AS (SELECT 1)
```

PR #329, opened 2021-02-04
Add "run_query" command

tvogels01:
This adds a new command `run_query` so that we see the result of queries underlying tables or views:
```
arthur.py run_query dw.fact
arthur.py run_query dw.fact --limit 10
```

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -55,6 +55,7 @@ from functools import partial
 from typing import Any, Dict, List, Optional, Set
 
 from psycopg2.extensions import connection  # only for type annotation
+from tabulate import tabulate
 
 import etl
 import etl.data_warehouse
@@ -1041,6 +1042,8 @@ def create_relations(
 
 # --- Section 5: "Callbacks" (functions that implement commands)
 
+# --- Section 5A: commands that modify tables and views
+
 
 def load_data_warehouse(
     all_relations: List[RelationDescription],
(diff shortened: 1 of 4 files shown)


python/etl/load.py
```diff
@@ -1202,6 +1205,26 @@ def update_data_warehouse(
...
+    dsn_etl = etl.config.get_dw_config().dsn_etl
+    query_stmt = relation.query_stmt + "\nLIMIT %s"
+
+    with Timer() as timer, closing(etl.db.connection(dsn_etl, autocommit=True)) as conn:
+        logger.info(
+            "Running query underlying '%s' (with 'LIMIT %s')",
+            relation.identifier,
+            limit if limit is not None else "NULL",
+        )
+        results = etl.db.query(conn, query_stmt, (limit,))
```
soumyadsanyal, line 1222: It seems like `conn.mogrify` handles the `LIMIT NULL` case if no format value is provided to `query_stmt`?

soumyadsanyal (approved):
👍

python/etl/load.py
tvogels01: The logging statement "manually" changes Python's `None` to SQL `NULL` so that the logging is consistent with what the user sees in the actual query run.
The function "mogrify" and psycopg2 in general translate between Python data types and SQL. So passing `None` into the query works fine and then shows up as `LIMIT NULL` in the query.

PR #330, opened 2021-02-04
Use "error" not just "warning" when relations fail to build

tvogels01:
When relations fail to build, then there's a stacktrace with a "critical" error only if that stops the ETL. If a non-required relation fails to build, this is only logged at the "warning" level. This makes these errors hard to find in the log output since there are so many other warnings.


soumyadsanyal (approved):
👍

PR #333, opened 2021-02-07
Add optional parameter for timeout to pizza pipeline

tvogels01:
This allows to run something like:
```
install_pizza_load_pipeline.sh development :transformations 8
```
in order to give the pipeline, say, 8 hours to complete instead of 6 (the default).

Side-effect: this reduces the diff nicely between `python/scripts/install_pizza_load_pipeline.sh` and  `python/scripts/install_rebuild_pipeline.sh`.


soumyadsanyal (approved):
👍

willcampbell433 (approved):
LGTM

PR #336, opened 2021-02-07
Fix message after failed retry

### LINKED ISSUE #162: Use library for retrying
# Summary

Use a library like [tenacity](https://github.com/jd/tenacity) instead of writing our own retrying logic.

# Details

That.

tvogels01:
We had a bug in the retry sequence where the tables that failed to extract were not listed in "keep going" mode. What we'd like to see is that we can run the extract pipeline (with `--keep-going`) and then see a list of failed relations which simplifies restarting the extract just for those relations that still need.

## Testing

One way to test this it to create a new settings file like this:
```
{
    "arthur_settings": {
        "extract_retries": 2
    }
}
```

Then "break" the extract by setting a bad password or connection and observe:

```
INFO etl.monitor (extract-source_0) [monitor.py:202] Starting extract step for '...' (1/3)
DEBUG etl.util.retry (extract-source_0) [retry.py:29] Starting on first of 3 attempt(s)
WARNING etl.util.retry (extract-source_0) [retry.py:43] Encountered the following error, but retrying 2 more time(s) after 1 second sleep: failed to fetch table size for '...'
DEBUG etl.util.retry (extract-source_0) [retry.py:31] There are now only 2 of 3 attempt(s) left
WARNING etl.util.retry (extract-source_0) [retry.py:43] Encountered the following error, but retrying 1 more time(s) after 5 second sleep: failed to fetch table size for '...'
DEBUG etl.util.retry (extract-source_0) [retry.py:31] There are now only 1 of 3 attempt(s) left
WARNING etl.monitor (extract-source_0) [monitor.py:226] Failed extract step for '...' (7.23s)
WARNING etl.extract.sqoop (extract-source_0) [extractor.py:105] Extract failed for non-required relation '...':
  File "/opt/src/arthur-redshift-etl/python/etl/extract/sqoop.py", line 56, in extract_table
    table_size = self.fetch_source_table_size(source.dsn, relation)
  File "/opt/src/arthur-redshift-etl/python/etl/extract/database_extractor.py", line 151, in fetch_source_table_size
    with closing(etl.db.connection(dsn_dict, readonly=True)) as conn:
  File "/opt/src/arthur-redshift-etl/python/etl/db.py", line 103, in connection
    cx = psycopg2.connect(**dsn_values)
  File "/opt/local/redshift_etl/venv/lib64/python3.7/site-packages/psycopg2/__init__.py", line 127, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  password authentication failed for user "etl"


The above exception was the direct cause of the following exception:

  File "/opt/src/arthur-redshift-etl/python/etl/util/retry.py", line 33, in call_with_retry
    func(*args, **kwargs)
  File "/opt/src/arthur-redshift-etl/python/etl/extract/sqoop.py", line 58, in extract_table
    raise DataExtractError("failed to fetch table size for '%s'" % relation.identifier) from exc
etl.errors.DataExtractError: failed to fetch table size for '...'

The above exception was the direct cause of the following exception:

  File "/opt/src/arthur-redshift-etl/python/etl/extract/extractor.py", line 100, in extract_table_with_retry
    call_with_retry(extract_retries, self.extract_table, source, relation)
  File "/opt/src/arthur-redshift-etl/python/etl/util/retry.py", line 37, in call_with_retry
    raise RetriesExhaustedError("reached max number of retries ({:d})".format(max_retries)) from exc
etl.errors.RetriesExhaustedError: reached max number of retries (2)

...

INFO etl.extract.sqoop (extract-source_0) [extractor.py:137] Finished extract from source '...': 0 succeeded, 3 failed (20.97s)
ERROR etl.extract.sqoop (MainThread) [extractor.py:159] Failed to extract from these source(s): '...'
WARNING etl.extract.sqoop (MainThread) [extractor.py:172] Failed to extract 3 relation(s): ...
```

## Other changes

* This PR starts a `etl.util` module with `retry` being the first inhabitant. Over time we'll move other supporting modules there.
* Raises minimum log level for `botocore.credentials` to `INFO`. (Lower again to `DEBUG` if you want to know the places that boto looks for credentials.

Closes #162 by cleaning up our code. A library would be nice but it would also take the logging away.


python/etl/relation.py
```diff
@@ -144,9 +144,10 @@ def load(self) -> None:
     @staticmethod
     def load_in_parallel(relations: List["RelationDescription"]) -> None:
         """Load all relations' table design file in parallel."""
+        max_workers = 8
+        logger.debug("Starting parallel load of %d table design file(s) on %d workers.", len(relations), max_workers)
         with etl.timer.Timer() as timer:
-            # TODO With Python 3.6, we should pass in a thread_name_prefix
-            with concurrent.futures.ThreadPoolExecutor(max_workers=8) as executor:
+            with concurrent.futures.ThreadPoolExecutor(max_workers=8, thread_name_prefix="load-parallel") as executor:
```
soumyadsanyal, line 150: `max_workers=8`

Should we just pass `max_workers=max_workers` as it's already set to 8 on L147?

tvogels01: Fixed. Thanks.

ynaim94 (approved):
LGTM, tested with a non required schema. Just one small observation/comment: If a user runs an `install_extraction_pipeline` on non required tables only (0 required table), if all of them fail to be extracted, the pipeline will still show `Finished`. Might be misleading if you don't think to go look at the logs.

tvogels01:
Yes, the pipeline would still be happy. 
1. We have `summarize_events` to see whether any `fail` events happened in the extract.
2. We could change the extract to return an error if any failures are found.

We don't want 2 to be the default so that the rebuild ETL continues even if a non-required table is missing. But we could add a command line option for the extraction pipeline.

PR #349, opened 2021-02-16
Make sure to create all groups when loading warehouse

tvogels01:
We can define groups in the configuration of the data warehouse and give them read or write access to schemas. This PR makes sure that all groups actually exist when that step is run for a rebuild ETL. You can also create the groups directly using the new `create_groups` command.

This does not allow to create arbitrary groups but only the ones referenced in the configuration file.

### DIFF
--- a/etc/arthur_completion.sh
+++ b/etc/arthur_completion.sh
@@ -17,6 +17,7 @@ _arthur_completion()
                 auto_design
                 bootstrap_sources
                 bootstrap_transformations
+                create_groups
                 create_index
                 create_schemas
                 create_user
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -583,7 +583,9 @@ class CreateGroupsCommand(SubCommand):
             "create_groups",
             "create all groups from the configuration",
             "Make sure that all groups mentioned in the configuration file actually exist."
-            " This allows to specify a group where the user is not also in the configuration.",
+            " (This allows to specify a group (as reader or writer) on a schema when that"
+            " group does not appear with a user and thus may not have been previously"
+            " created using a 'create_user' call.)",
         )
 
     def add_arguments(self, parser):
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -293,6 +293,7 @@ def build_full_parser(prog_name):
         # Commands to deal with data warehouse as admin:
         InitializeSetupCommand,
         ShowRandomPassword,
+        CreateGroupsCommand,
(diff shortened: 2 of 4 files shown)


ynaim94 (commented):
Code looks good, but:
```
$ arthur.py create_groups youssef_test                
usage: arthur.py [-h] [-c CONFIG | --submit-to-cluster CLUSTER_ID] [-V]
                 {initialize,show_random_password,create_groups,create_user,update_user,bootstrap_sources,design,bootstrap_transformations,auto_design,validate,explain,run_query,show_ddl,sync,extract,load,upgrade,update,unload,create_schemas,promote_schemas,ping,terminate_sessions,ls,show_downstream_dependents,show_dependents,show_upstream_dependencies,render_template,show_value,show_vars,settings,show_pipelines,delete_finished_pipelines,query_events,summarize_events,tail_events,help,selftest}
                 ...
arthur.py: error: unrecognized arguments: youssef_test
```
does not seem correct. 

1 - It is supposed to give an error message related to the command.

```
 $ arthur.py create_groups -h
usage: arthur.py create_groups [-h] [-o | -v | -q] [-n]

Make sure that all groups mentioned in the configuration file actually exist.
This allows to specify a group where the user is not also in the
configuration.

optional arguments:
  -h, --help     show this help message and exit
  -o, --prolix   send full log to console
  -v, --verbose  increase verbosity
  -q, --quiet    decrease verbosity
  -n, --dry-run  do not modify stuff
```
I am not allowed to give groups as argument.

Am I missing something?

tvogels01:
Regarding 1:

> arthur.py: error: unrecognized arguments: youssef_test

That's the output regarding the error. The command didn't recognize the additional argument.

Regarding 2:

Ok, I see where the help text is misleading.

> This allows to specify a group where the user is not also in the configuration.

The "specify" refers to the configuration file, not the command line. 

Overall, the goal is to make sure that any group mentioned in a configuration file already exists. The goal is not to create groups outside of that so you can't specify anything on the command line.

Let's say we want to define that a new group of users `ax_ro` who have access to our star schema:
```
            {
                "name": "data_mart",
                "description": "Read-optimized dimensional models (star schemas)",
                "groups": [ "etl_ro", "ax_ro" ]
            }
```

This would break before the patch because no group `ax_ro` exists but the ETL would still try to grant permissions to that group.
Next step is to add a `groups` section in the configuration file to make sure that groups are always declared so that spelling mistakes don't creep in. (`ax_ro` vs. `axro` etc.)

PR #351, opened 2021-02-20
Use pre-commit

tvogels01:
This gets rid of the home-grown pre-commit hook and instead uses the
[`pre-commit`](https://pre-commit.com/) project.

Result of committing the new configuration:
```shell
% pre-commit run --all      
Check for broken symlinks................................................Passed
Fix End of Files.........................................................Passed
black....................................................................Passed
flake8...................................................................Passed
isort....................................................................Passed
mypy.....................................................................Passed
pyupgrade................................................................Passed
```

The current configuration with `requirements-linters.txt` is left in place to allow continued use, _e.g._ from within an editor.

This PR also switches YAML files to the common 2-space indentation but doesn't update the existing files. We'll update them whenever we touch them again.

Installation: see README.md changes


willcampbell433 (approved):
I see mostly formatting updates and then swapping out the `pre-commit` project instead of the homegrown one --- aka. LGTM

PR #357, opened 2021-02-27
Rename "pizza load" to "pizza" pipeline

tvogels01:
## Description

This renames our "pizza load" pipeline to "pizza" pipeline since the former is a bit of a misnomer in this case:
```
install_pizza_load_pipeline.sh development ":transformations"
```
So this now highlights that truly only transformations are run:
```
install_pizza_pipeline.sh development ":transformations"
```

Although this is a "rename", we leave the existing script in place so that current deploy instructions don't break. We can drop the "pizza load" version with v2.

## Bug fixes

The timeout parameter is now actually picked up and set on the data pipeline.

## Developer
* Adding `shellcheck` here and cleaning up the pizza pipeline script accordingly.

## Testing

For testing, it might be safest to comment out the `activate-pipeline` step at the bottom of the script.


ynaim94:
> Bug fixes
> The timeout parameter is now actually picked up and set on the data pipeline.

This did not seem to work based on the pizza pipeline you deployed this morning with a timeout of 10h. It was still setup to 6 in the UI.

tvogels01:
>> Bug fixes
>> The timeout parameter is now actually picked up and set on the data pipeline.

> This did not seem to work based on the pizza pipeline you deployed this morning with a timeout of 10h. It was still setup to > 6 in the UI.

Here's what I started:
```
install_pizza_pipeline.sh tom common.numbers 13
```

Then I checked the parameters
```
                        "@Parameters": {
                            "name": null,
                            "ref_values": {},
                            "string_values": {
                                "@sphere": "COMPONENT",
                                "@status": "PENDING",
                                "@version": "1",
                                "mySelection": "common.numbers",
                                "myStartDateTime": "2021-03-01T17:49:24",
                                "myTimeout": "13"
                            }
                        },
```

The problem earlier this morning was related to setting tags.

ynaim94 (approved):
LGTM

PR #363, opened 2021-03-02
Set AWS_REGION based on VPC region

tvogels01:
Data Pipeline, unlike, say, Lambda, doesn't set the region for resources that it starts. So we're missing this information and trying to get a client errors out. This PR sets the `AWS_DEFAULT_REGION` based on `resources.VPC.region`.

This is the bugfix for:
```
|  botocore.exceptions.NoRegionError: You must specify a region.
```

Other changes:
* Update the type declarations

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -119,21 +119,29 @@ def run_arg_as_command(my_name="arthur.py"):
         with execute_or_bail():
             etl.config.load_config(args.config)
 
-            setattr(args, "bucket_name", etl.config.get_config_value("object_store.s3.bucket_name"))
             if hasattr(args, "prefix"):
+                # Any command where we can select the "prefix" also needs the bucket.
+                # TODO(tom): Need to differentiate between object store (schemas) and data lake
+                #     (extracted or unloaded data)
+                setattr(args, "bucket_name", etl.config.get_config_value("object_store.s3.bucket_name"))
                 etl.config.set_config_value("object_store.s3.prefix", args.prefix)
                 etl.config.set_config_value("data_lake.s3.prefix", args.prefix)
+
                 # Create name used as prefix for resources, like DynamoDB tables or SNS topics
                 base_env = etl.config.get_config_value("resources.VPC.name").replace("dw-vpc-", "dw-etl-", 1)
                 etl.config.set_safe_config_value("resource_prefix", "{}-{}".format(base_env, args.prefix))
+
                 if getattr(args, "use_monitor"):
                     etl.monitor.start_monitors(args.prefix)
+
+            # The region must be set for most boto3 calls to succeed.
+            os.environ["AWS_REGION"] = etl.config.get_config_value("resources.VPC.region")
 
             dw_config = etl.config.get_dw_config()
             if isinstance(getattr(args, "pattern", None), etl.names.TableSelector):
                 args.pattern.base_schemas = [s.name for s in dw_config.schemas]
 
-            # TODO Remove dw_config and let sub-commands handle it!
+            # TODO(tom): Remove dw_config and let sub-commands handle it!
             args.func(args, dw_config)
 
 
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -135,7 +135,7 @@ def run_arg_as_command(my_name="arthur.py"):
                     etl.monitor.start_monitors(args.prefix)
 
             # The region must be set for most boto3 calls to succeed.
-            os.environ["AWS_REGION"] = etl.config.get_config_value("resources.VPC.region")
+            os.environ["AWS_DEFAULT_REGION"] = etl.config.get_config_value("resources.VPC.region")
 
             dw_config = etl.config.get_dw_config()
             if isinstance(getattr(args, "pattern", None), etl.names.TableSelector):
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -33,8 +33,8 @@ logger.addHandler(logging.NullHandler())
 
 
 # Global config objects - always use accessors!
-_dw_config = None  # type: Optional[DataWarehouseConfig]
-_mapped_config = None  # type: Optional[Dict[str, str]]
+_dw_config: Optional[DataWarehouseConfig] = None
+_mapped_config: Optional[Dict[str, str]] = None
 
 # Local temp directory used for bootstrap, temp files, etc.
 # TODO(tom): This is a misnomer -- it's also the install dir on EC2 hosts.
@@ -258,7 +258,7 @@ def load_config(config_files: Sequence[str], default_file: str = "default_settin
 
     The settings are validated against their schema.
     """
-    settings = dict()  # type: Dict[str, Any]
+    settings: Dict[str, Any] = dict()
     count_settings = 0
     for filename in yield_config_files(config_files, default_file):
         if filename.endswith(".sh"):
@@ -271,7 +271,7 @@ def load_config(config_files: Sequence[str], default_file: str = "default_settin
 
     # Need to load at least the defaults and some installation specific file:
     if count_settings < 2:
-        raise ETLRuntimeError("Failed to find enough configuration files (need at least default and local config)")
+        raise ETLRuntimeError("failed to find enough configuration files (need at least default and local config)")
 
     validate_with_schema(settings, "settings.schema")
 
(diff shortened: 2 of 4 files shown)


python/etl/commands.py
```diff
@@ -119,21 +119,29 @@ def run_arg_as_command(my_name="arthur.py"):
...
+
                 # Create name used as prefix for resources, like DynamoDB tables or SNS topics
                 base_env = etl.config.get_config_value("resources.VPC.name").replace("dw-vpc-", "dw-etl-", 1)
                 etl.config.set_safe_config_value("resource_prefix", "{}-{}".format(base_env, args.prefix))
+
                 if getattr(args, "use_monitor"):
                     etl.monitor.start_monitors(args.prefix)
 
+            # The region must be set for most boto3 calls to succeed.
+            os.environ["AWS_REGION"] = etl.config.get_config_value("resources.VPC.region")
```
ynaim94, line 138: I feel like we might open a possibility for user misconfiguration and inconsistency? What about a solution like this: https://stackoverflow.com/questions/4249488/find-region-from-within-an-ec2-instance

python/etl/config/__init__.py
```diff
@@ -33,8 +33,8 @@
 
 
 # Global config objects - always use accessors!
-_dw_config = None  # type: Optional[DataWarehouseConfig]
-_mapped_config = None  # type: Optional[Dict[str, str]]
+_dw_config: Optional[DataWarehouseConfig] = None
```
ynaim94, line 36: Curious, why changing these? Linter complaining?

python/etl/pipeline.py
```diff
@@ -91,15 +91,18 @@ def status(self):
...
-    client = boto3.client("datapipeline")
-
     def __init__(self, description) -> None:
         self.name = description["name"]
         self.pipeline_id = description["pipelineId"]
         self.string_values = {
             field["key"]: field["stringValue"] for field in description["fields"] if "stringValue" in field
         }
         self.tags = description["tags"]
+        self.client = boto3.client("datapipeline")
```
ynaim94, line 101: If I'm understanding correctly, you moved this to an instance variable to insure that `AWS_REGION` is setup before instantiating a pipeline (as I believe that's what boto3 use by default if there is no `~/aws` config).  Question:
1 - Can a data pipeline ever be instantiated outside an Arthur command?
2 - Could we make the dependency more explicit by adding it the region to the constructor?

ynaim94 (commented):
Back to you with a few questions

python/etl/commands.py
tvogels01: Take a look at what the bootstrap script does. It checks in with the API for instance values. Theoretically we would know at that point and could store that configuration. That's if you want to follow stackoverflow and "discover" the region.

But I'm not sure where the inconsistency would come from though. If the region isn't set correctly, how would the rest of the data pipeline work?
https://github.com/harrystech/arthur-redshift-etl/blob/v1.38.0/python/etl/templates/ec2_instance.json#L17
https://github.com/harrystech/arthur-redshift-etl/blob/v1.38.0/python/etl/templates/rebuild_pipeline.json#L12

We already rely on this setting here:
https://github.com/harrystech/arthur-redshift-etl/blob/v1.38.0/python/etl/monitor.py#L305

At the moment, we can't really support multi-region deployment. E.g. you can't bring up an EMR cluster in us-west-2 while running Redshift in us-west-1. We should tackle that but that would blow up the scope of this PR.

python/etl/config/__init__.py
tvogels01: https://github.com/harrystech/arthur-redshift-etl/blob/v1.39.0-rc3/.pre-commit-config.yaml#L28

We've upgraded from Python 3.5 so this brings in language changes from Python 3.6.

python/etl/pipeline.py
tvogels01: I'd say the alternative might be keep the client outside of the class, like `def describe_pipelines(client, pipeline_ids: List[str])` does.  I've changed to that model.

Boto3 uses `AWS_DEFAULT_REGION` or `AWS_REGION` even when there is a `~/.aws/config`. We should be able to rely on the environment variables. The question now is how do we set that env. variable in a simple way? Given that we don't have multi-region support in this setup, a central place seems like a fair place to do so.

If somebody else wants to use `etl.pipeline`, then they can do so. They may have to set the `region_name` on the client, depending on the configuration, but that's normal for working with boto3.

tvogels01:
This works fine on the EC2 instance that's launched:
```
/tmp/redshift_etl/venv/bin/arthur.py -c /tmp/redshift_etl/config/ show_pipelines
```

ynaim94:
@tvogels01 Just launched an EC2 instance, testing:
```
$ /tmp/redshift_etl/venv/bin/arthur.py -c /tmp/redshift_etl/config/ show_pipelines
2021-03-02 18:12:12 - CRITICAL - Something terrible happened:
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/commands.py", line 76, in execute_or_bail
    yield
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/commands.py", line 137, in run_arg_as_command
    args.func(args, dw_config)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/commands.py", line 1369, in callback
    etl.pipeline.show_pipelines(args.selection)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/pipeline.py", line 77, in show_pipelines
    pipelines = list_pipelines(selection)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/pipeline.py", line 48, in list_pipelines
    client = boto3.client("datapipeline")
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/boto3/__init__.py", line 93, in client
    return _get_default_session().client(*args, **kwargs)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/session.py", line 838, in create_client
    client_config=config, api_version=api_version)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/client.py", line 87, in create_client
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/client.py", line 328, in _get_client_args
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/args.py", line 73, in get_client_args
    endpoint_url, is_secure, scoped_config)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/args.py", line 154, in compute_client_args
    s3_config=s3_config,
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/args.py", line 220, in _compute_endpoint_config
    return self._resolve_endpoint(**resolve_endpoint_kwargs)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/args.py", line 303, in _resolve_endpoint
    service_name, region_name, endpoint_url, is_secure)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/client.py", line 402, in resolve
    service_name, region_name)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/regions.py", line 134, in construct_endpoint
    partition, service_name, region_name)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/botocore/regions.py", line 148, in _endpoint_for_partition
    raise NoRegionError()
botocore.exceptions.NoRegionError: You must specify a region.
2021-03-02 18:12:12 - INFO - Ran for 0.46s before encountering disaster!
Bailing out: botocore.exceptions.NoRegionError: You must specify a region.
```
Am I doing something wrong?

tvogels01:
@ynaim94 Are you sure that you were running the latest code from this branch? (6305472)

```
bin/build_arthur.sh
bin/run_arthur.sh
```
```
md5sum /opt/src/arthur-redshift-etl/python/etl/pipeline.py
# c04b5accc143adc83848125dcdffc8ef  /opt/src/arthur-redshift-etl/python/etl/pipeline.py
upload_env.sh

launch_ec2_instance.sh
# ssh per output from script
```
```
md5sum /tmp/redshift_etl/venv/lib/python3.7/site-packages/etl/pipeline.py
# c04b5accc143adc83848125dcdffc8ef  /tmp/redshift_etl/venv/lib/python3.7/site-packages/etl/pipeline.py

/tmp/redshift_etl/venv/bin/arthur.py -c /tmp/redshift_etl/config/ show_pipelines
# 2021-03-03 11:05:06 - INFO - Starting log for redshift_etl v1.39.0 with ETL ID 8EAE3754E39F4216
```

ynaim94 (approved):
The upload env got me again. LGTM

PR #366, opened 2021-03-06
Cleanup monitor commands

tvogels01:
Only internal changes to simplify the code.
* Leverage abstract base class
* Move the default attribute for `uses_monitor` (as `False`) into that class.

### Testing

Start container with exposing port 8086.
```
bin/run_arthur.sh -w
```
Start the demo in the container:
```
python -m etl.monitor
```

And then follow along: http://localhost:8086/

### Monitor events

Make sure that the ETL events are still sent to the DynamoDB events table for extract, load, update, upgrade, and unload.

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -132,6 +132,7 @@ def run_arg_as_command(my_name="arthur.py"):
     if args.cluster_id is not None:
         submit_step(args.cluster_id, args.sub_command)
         return
+<<<<<<< HEAD
 
     # We need to configure logging before running context because that context expects
     # logging to be setup.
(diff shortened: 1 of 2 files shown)


python/etl/commands.py
```diff
@@ -890,7 +885,7 @@ def add_arguments(self, parser):
         )
         group.add_argument(
             "--use-existing-csv-files",
-            help="skip extraction and go straight to creating manifest files, implied default for static sources",
+            help="skip extraction and go straight to creating manifest files," " implied default for static sources",
```
tvogels01, line 888: These broken strings will go onto two lines when we reduce max line width to 100.

PR #379, opened 2021-03-17
Filter the "AUTO" setting for column encoding

tvogels01:
We currently allow the setting "AUTO" in a settings file to pick an automatic encoding of columns in Redshift tables. But that feature isn't complete. This PR prevents this setting from accidentally being called an breaking the `COPY` statement (with a  `COMPUPDATE AUTO`) by replacing `AUTO` with `ON` (which is ok in `COMPUPDATE ON`.
As before, if encodings are specified for all columns, "compupdate" is turned off.


soumyadsanyal (approved):
👍

PR #380, opened 2021-03-18
Show parameters of refresh pipeline

tvogels01:
This PR adds the pipeline parameters (for selection of relations) to the success and failure notifications. This makes it easier to re-run refresh pipelines by looking up the parameter in the email, which is unlike today when we have to look up parameters in deployment notices or spreadsheets.
(Although the driver here is refresh, the parameter is also shown for the pizza pipeline.)

Example output in a message:
```
Last node: @Send Health Check After ETL (EC2)_2021-03-18T12:31:27_Attempt=1
PipelineId: df-08176362XEI8C3GGYA1D
Parameters: web_app.orders
Log directory: s3://harrys-data-warehouse-dev/_logs/tom/current/df-08176362XEI8C3GGYA1D
```


soumyadsanyal (approved):
👍

PR #420, opened 2021-04-27
Add test for distribution key

tvogels01:
Turns out the validation for table designs wasn't making sure that a distribution key for a table is a column in that table. This PR changes the validation for table attributes and constraints so that this is addressed by taking a more general approach: all table attributes and constraints that aren't tokens (like `ALL` for a distribution key) will have the list of columns checked against the list of columns present in the table.

### DIFF
--- a/python/etl/design/load.py
+++ b/python/etl/design/load.py
@@ -9,8 +9,9 @@ Table designs are implemented as dictionaries of dictionaries or lists etc.
 
 import logging
 from contextlib import closing
+from typing import Dict, List
 
-import funcy as fy
+import funcy
 import yaml
 import yaml.parser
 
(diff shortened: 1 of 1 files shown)


swong28 (approved):
LGTM. it is giving me errors when running `arthur.py bootstrap_transformations update analytics.zendesk_ticket_attributes` now!

ynaim94:
@tvogels01 is the use case when a column is present in one of the parent table is used as a attribute/constraint, but is not present in the child table? Because I'm testing with the current version of arthur and it is complaining 
```shell
etl.errors.TableDesignSemanticError: primary_key columns in constraints contain unknown column(s): ...
```

ynaim94 (approved):
LGTM

Compared master and this PR on validating a table design where the distribution key was not in the final columns:
* Master: Didn't complain.
* This PR: Errored out with  `etl.errors.TableDesignSemanticError: unknown distribution column(s) in table attributes: ...`

PR #426, opened 2021-05-09
Add option to skip loading sources (so "ELT" is "T" only)

tvogels01:
This PR adds the `--skip-loading-sources` option to the `load` command:
```
arthur.py load --skip-loading-sources
```

The goal here is to run all transformations on empty sources tables. This allows us to run through all the transformation queries, which thus get compiled before the full ELT needs them. 

Since we cannot leave external tables empty, which are controlled by AWS Glue Data catalog, we also skip transformations that depend on an external table. We can identify those by not being managed by Arthur.

* https://aws.amazon.com/about-aws/whats-new/2020/06/amazon-redshift-now-delivers-better-cold-query-performance/
* https://aws.amazon.com/blogs/big-data/fast-and-predictable-performance-with-serverless-compilation-using-amazon-redshift/

### DIFF
--- a/python/etl/dialect/redshift.py
+++ b/python/etl/dialect/redshift.py
@@ -209,7 +209,7 @@ def build_view_ddl(view_name: TableName, columns: List[str], query_stmt: str) ->
 
 def build_insert_ddl(table_name: TableName, column_list, query_stmt) -> str:
     """Assemble the statement to insert data based on a query."""
-    columns = join_with_double_quotes(column_list)
+    columns = join_with_double_quotes(column_list, sep=",\n            ")
     insert_stmt = """
         INSERT INTO {table} (
             {columns}
--- a/python/etl/dialect/redshift.py
+++ b/python/etl/dialect/redshift.py
@@ -432,6 +432,9 @@ def insert_from_query(
         logger.info("Inserting data into '%s' from query", table_name.identifier)
         try:
             etl.db.execute(conn, stmt)
+        # TODO(tom) Ideally we'd retry these but initial testing ran into issues with closed connections.
+        # except psycopg2.OperationalError as exc:
+        #    raise TransientETLError(exc) from exc
         except psycopg2.InternalError as exc:
             if exc.pgcode in retriable_error_codes:
                 raise TransientETLError(exc) from exc
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -114,11 +114,10 @@ class LoadableRelation:
         """Grab everything from the contained relation. Fail if it's actually not available."""
         if hasattr(self._relation_description, name):
             return getattr(self._relation_description, name)
-        raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, name))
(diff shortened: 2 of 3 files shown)


python/etl/dialect/redshift.py
```diff
@@ -423,9 +423,11 @@ def insert_from_query(
         logger.info("Inserting data into '%s' from query", table_name.identifier)
         try:
             etl.db.execute(conn, stmt)
+        except psycopg2.OperationalError as exc:
```
tvogels01, line 426: Maybe revert catching `OperationalError` and have that as a separate PR.

python/etl/load.py
```diff
@@ -260,8 +258,18 @@ def from_descriptions(
...
         base_destination = {"name": database}
 
+        managed = frozenset(relation.identifier for relation in relations)
+
         loadable = []
         for i, relation in enumerate(relations):
+            this_skip_copy = skip_copy
+            if skip_loading_sources:
+                # Only load transformations and only if no dependency is external.
+                if not relation.is_transformation or any(
```
ynaim94, line 268: When is a relation not a transformation? I see that it's when `self.kind == "DATA"`, when does that happen?

ynaim94 (approved):
One thing I'm confused about but otherwise LGTM.

python/etl/load.py
tvogels01: This PR will have some updates that will help to distinguish the relations: https://github.com/harrystech/arthur-redshift-etl/pull/202/files

Transformations are CTAS or view relations. A table loaded from a source is not considered a transformation. Practically it simply means that it's a transformation if (and only if) you have a `.sql` file.

python/etl/load.py
```diff
@@ -700,27 +708,25 @@ def build_one_relation(conn: connection, relation: LoadableRelation, dry_run=Fal
...
-                rowcount = rows[0]["rowcount"]
-                logger.info("Found {:d} row(s) in {:x}".format(rowcount, relation))
-                monitor.add_extra("rowcount", rowcount)
+        if relation.is_view_relation or relation.failed or relation.skip_copy:
+            return
+        stmt = f"SELECT COUNT(*) AS rowcount FROM {relation}"
+        if dry_run:
+            etl.db.skip_query(conn, stmt)
+            return
+        rows = etl.db.query(conn, stmt)
```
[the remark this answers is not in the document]
tvogels01: Yes, there's overall some code cleanup going on. In general you'll find me favoring:
```
if blah:
    do_x()
    return

do_y()
```
over
```
if blah:
    do_x()
else:
    do_y()
```

The idea is to take care of "abnormal" branches and then make sure that the expected (or happy path) isn't indented too deeply.

So in this case, we ended up with the twice-indented info about row count 🤷 
But mostly, this allowed to have a cleaner (or is it?) description of the query that's running to get the row count.

Finally (and maybe I'm wrong here?), this will skip the row count for tables that were left empty. This buys us some time in validation and cache warmup.

PR #450, opened 2021-06-24
Add CloudWatch logging to ETL

willcampbell433:
This PR allows the ETL to log to Cloudwatch by using the [Watchtower](https://github.com/kislyuk/watchtower) log handler.

Configuration:
* Only if enabled in the configuration will logs be sent to CloudWatch. (So you have to opt-in.)
* The log group is set in the configuration and applies globally.
    * The log group will be created if it doesn't already exist.
    * The retention period will be set to 6 months.
* The log stream is derived from the prefix (environment) and the date.

You can then check logs using the new `tail_logs` command.

Known issues:
* Any subscription filters (e.g. to forward logs into log collection) must be manually set up.
* Some commands do not use a prefix argument and logs will show up as `ec2-user`.
* When running any command, logs will actually be sent to CloudWatch (when enabled). -- We should reduce this to just the "monitored" commands.

---

In order to test the logging to CloudWatch, you need to enable it.
Copy this into a new settings file and deploy into your `config` directory:

```yaml
{
  "arthur_settings": {
    "logging": {
      "cloudwatch": {
        "enabled": true
      }
    }
  }
}
```

---

In order for the EC2 host to send logs to CloudWatch, we're widening the permissions for the role that the EC2 hosts assume.

You will have to update the roles, e.g. by creating a change set and then executing it:

```shell
cloudformation/do_cloudformation.sh create-change-set dw_vpc dev \
    ObjectStore=UsePreviousValue \
    DataLake=UsePreviousValue \
    SecondaryDataLake=UsePreviousValue \
    VpcNetworkSelection=UsePreviousValue \
    WhitelistCIDR1=UsePreviousValue \
    WhitelistCIDR2=UsePreviousValue \
    WhitelistCIDR3=UsePreviousValue \
    WhitelistCIDR4=UsePreviousValue \
    WhitelistCIDR5=UsePreviousValue
```
(be sure to use the appropriate qualifier instead of `dev`)


willcampbell433:
After review with @tvogels01 and walking through features, LGTM. Let's ship to dev and let it run over the weekend.

tvogels01 (approved):
LGTM

PR #463, opened 2021-07-07
DENG-822: Empty run pipeline

ynaim94:
## Description

Creating a script to deploy a "Warmup" pipeline, which allows us to load the data with the `--skip-loading-sources` argument. This has for goal to warmup the pipeline by running all the SQL transformations and pre-compiling them before the post maintenance ETL Rebuild Pipeline that occurs in Saturday evening.

## Jira ticket

https://harrys.atlassian.net/browse/DENG-822

## QA

Installing the pipeline and running it:

```SHELL
(aws:data-dev, prefix:youssef) $ install_warmup_pipeline.sh youssef 2021-07-14T22:00:00 10
```

[Link to the pipeline in the AWS Console](https://console.aws.amazon.com/datapipeline/?#ExecutionDetailsPlace:pipelineId=df-0083505S7CU6U7JFPZX&show=latest)


tvogels01:
1. What would this look like if it were more patterned after the validation pipeline (so live as `production/post_maintenance`? leverage the same script for creating credentials?)
2. What would it look like if we made this part of the validation? (So instead of skipping all queries only skip sources? Maybe as an option?)
3. This doesn't need the code for the changes of `load` in the PR. (It'll complicate your live when those changes are merged.)
4. Any feedback or ideas about names? ("empty" may not help us much in the long run.)

ynaim94:
> 2. What would it look like if we made this part of the validation? (So instead of skipping all queries only skip sources? Maybe as an option?)

Wouldn't that make the validation pipeline a lot longer? 
> 3. This doesn't need the code for the changes of load in the PR. (It'll complicate your live when those changes are merged.)

Yes I will revert some of the changes there.
> 4. Any feedback or ideas about names? ("empty" may not help us much in the long run.)

I'm the one that chose the name I think 😅 I'm the worst at picking up names, do we want something more purpose based like post-maintenance pipeline?

tvogels01:
>> What would it look like if we made this part of the validation? (So instead of skipping all queries only skip sources? Maybe as an option?)
>> Wouldn't that make the validation pipeline a lot longer?

Yes, it would. But I'm not suggesting to change the normal validation but instead add an option to the existing tone.

* Note that `validation_ ...` databases are easily supported with the `initialize` command. So using `validation_no_load` would be easier than another name like `staging`.
* Re-using the pipeline would reduce additional code. We already have a bunch of "install" scripts for pipelines. (See #462 ... lots of scripts.)

>> This doesn't need the code for the changes of load in the PR. (It'll complicate your live when those changes are merged.)
> Yes I will revert some of the changes there.

Thanks. Let's accelerate the other PR (#426).

>> Any feedback or ideas about names? ("empty" may not help us much in the long run.)
> I'm the one that chose the name I think 😅 I'm the worst at picking names, do we want something more purpose based like post-maintenance pipeline?

Practice :-) Something like either "post-restart" or "warm-up" would work in my mind. This is really about warming up the query cache and potentially for us to actually test the queries live.

tvogels01:
Also, if this lands as a separate pipeline first and then we'll rethink how we can refactor pipelines, that's ok.

tvogels01 (commented):
Looks like I never submitted the review after our discussion last week :(

bin/create_validation_credentials
```diff
@@ -2,16 +2,19 @@
...
 
+DEFAULT_VALIDATION_FOLDER="validation"
+
 set -o errexit -o nounset
 
 show_usage_and_exit() {
   cat <<USAGE
 
-Usage: $(basename "$0") <bucket_name> <prefix>
+Usage: $(basename "$0") <bucket_name> <prefix> <validation-folder>
```
tvogels01, line 12: Since this is an optional argument, it should be set within `[ ... ]`.

```suggestion
Usage: $(basename "$0") <bucket_name> <prefix> [<validation-folder>]
```

bin/create_validation_credentials
```diff
@@ -2,16 +2,19 @@
...
 
 This will create an additional credentials file needed for the validation pipeline.
 By changing the target database from "database" to "validation_database", we can make sure that
 the validation pipeline runs on an empty and separate database.
+The validation folder name defaults to "validation"
```
tvogels01, line 17: I think that it would be safer to refer back to what sets the default and also to give a bit of an explanation.

```suggestion

The validation folder name defaults to "$DEFAULT_VALIDATION_FOLDER".
That is the folder under the environment in "prefix" to be used for the validation setup.
```

bin/create_validation_credentials
```diff
@@ -23,13 +26,14 @@ case "${1-help}" in
     ;;
 esac
 
-if [[ $# -ne 2 ]]; then
+if [[ $# -lt 2 ]]; then
```
tvogels01, line 29: It's safer to check for the range, not just the lower bound.

```suggestion
if [[ $# -lt 2 || $# -gt 3 ]]; then
```

bin/create_validation_credentials
```diff
@@ -23,13 +26,14 @@ case "${1-help}" in
...
   echo 1>&2 "Wrong number of arguments."
   show_usage_and_exit 1
 fi
 
 PROJ_BUCKET="$1"
 PROJ_ENVIRONMENT="$2"
+VALIDATION_FOLDER="${3:-DEFAULT_VALIDATION_FOLDER}"
```
tvogels01, line 36: This is missing a `$` so would use the variable name not variable value.

```suggestion
VALIDATION_FOLDER="${3:-$DEFAULT_VALIDATION_FOLDER}"
```

python/scripts/install_warmup_pipeline.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
+
+START_NOW=$(date -u +"%Y-%m-%dT%H:%M:%S")
+USER="${USEmR-nobody}"
```
tvogels01, line 4: Careful :-)

```suggestion
USER="${USER-nobody}"
```

python/scripts/install_warmup_pipeline.sh
```diff
@@ -0,0 +1,95 @@
...
+DEFAULT_PREFIX="${ARTHUR_DEFAULT_PREFIX-$USER}"
+
+if [[ $# -gt 3 || "$1" = "-h" ]]; then
+    cat <<USAGE
```
tvogels01, line 8: We'll be moving to 2 spaces. That's ok to keep here since those changes went into `next` after you started.

https://google.github.io/styleguide/shellguide.html#indentation

python/scripts/install_warmup_pipeline.sh
```diff
@@ -0,0 +1,95 @@
...
+USAGE
+    exit 0
+fi
+
+set -o errexit -o nounset
+
+# Verify that there is a local configuration directory
+DEFAULT_CONFIG="${DATA_WAREHOUSE_CONFIG:-./config}"
+if [[ ! -d "$DEFAULT_CONFIG" ]]; then
+    echo "Failed to find \'$DEFAULT_CONFIG\' directory."
```
tvogels01, line 27: We should be sending this to stderr with `1>&2`. (Maybe next iteration.)

To compare 🤷 
```
git co next python/scripts/install_validation_pipeline.sh
diff python/scripts/install_validation_pipeline.sh python/scripts/install_warmup_pipeline.sh
```

bin/create_validation_credentials
```diff
@@ -23,13 +26,14 @@ case "${1-help}" in
...
 
 # Figure out the "safe" name that drops suspicious characters (replacing them with '_').
 SAFE_ENVIRONMENT="$(echo -n "$PROJ_ENVIRONMENT" | tr -cs '_a-zA-Z0-9' '_')"
```
tvogels01, line 39: The target database should probably be derived from the folder name, so `validation_warmup`.

```suggestion
SAFE_ENVIRONMENT="$(echo -n "$PROJ_ENVIRONMENT" | tr -cs '_a-zA-Z0-9' '_')"
SAFE_VALIDATION_FOLDER="$(echo -n "$VALIDATION_FOLDER" | tr -cs '_a-zA-Z0-9' '_')"
VALIDATION_DATABASE="${SAFE_VALIDATION_FOLDER}_${SAFE_ENVIRONMENT}"
```

bin/create_validation_credentials
```diff
@@ -23,13 +26,14 @@ case "${1-help}" in
     ;;
 esac
 
-if [[ $# -ne 2 ]]; then
+if [[ $# -lt 2 ]]; then
   echo 1>&2 "Wrong number of arguments."
   show_usage_and_exit 1
 fi
```
tvogels01, line 32: Maybe check name of folder?

This script is called the "create_validation_credentials" -- so it seems fair that we always use a "validation_*" database.

```suggestion
fi

CASE
  valiadation_*) true;;
  *)
      echo 1>&2 "Bad folder name, should start with 'validation_'
      exit 1
      ;;
ESAC
```

tvogels01 (commented):
Some suggestions.

python/scripts/install_warmup_pipeline.sh
```diff
@@ -0,0 +1,95 @@
...
+aws datapipeline create-pipeline \
+    --unique-id warmup-pipeline \
+    --name "$PIPELINE_NAME" \
+    --tags $AWS_TAGS \
+    | tee "$PIPELINE_ID_FILE"
+
+PIPELINE_ID=$(jq --raw-output < "$PIPELINE_ID_FILE" '.pipelineId')
+
+if [[ -z "$PIPELINE_ID" ]]; then
+    set +x
```
tvogels01, line 78: One more thing :-)

It's nicer to maintain when this is explicit so 
```suggestion
  set +o xtrace
```
But also, these are changes just coming in recently so ok to omit on this PR.

ynaim94:
@tvogels01 I addressed all the comments except the ones regarding using mirroring the new validation script which I will do in a following iteration.

ynaim94:
Running a pipeline here : https://console.aws.amazon.com/datapipeline/home?region=us-east-1#ExecutionDetailsPlace:pipelineId=df-05087381GONMESXIACG3&show=latest

tvogels01 (approved):
LGTM

PR #474, opened 2021-07-14
Allow to specify start time when searching logs

tvogels01:
This will filter logs from CloudWatch (assuming they were enabled) by log stream and timestamp.

Usage:
```
arthur.py tail_logs

arthur.py tail_logs -t 2021-07-14T10:29:33
```

This allows us to quickly review what "just happened" when we get alerted for an incident.


willcampbell433:
Tested with `arthur.py tail_logs -t 2021-07-14T10:29:33` and was able to search for logs given a timestamp
image

willcampbell433 (approved):

PR #491, opened 2021-08-06
Fix comma-separated selection in refresh pipeline

tvogels01:
Fixes a bug where the list of tables was no longer comma-separated when passed to the EMR cluster. (This affected the install script for Refresh ETLs.)


willcampbell433 (approved):
LGTM

PR #492, opened 2021-08-06
Add exception class and message to CloudWatch

tvogels01:
This changes the exception information to split into class and message like so:
```
"exception": {
    "class": "UpdateTableError",
    "message": "Cannot insert a NULL value into column site
DETAIL:  
  -----------------------------------------------
  error:  Cannot insert a NULL value into column site
  code:      8007
  context:   query execution
  query:     5497726
  location:  column:1
  process:   query0_102_5497726 [pid=20385]
  -----------------------------------------------

"
}
```
This allows easy search for the exception class and cleans up the message.


willcampbell433 (approved):
LGTM

PR #523, opened 2021-09-15
Test on number of execution levels

tvogels01:
There's an error in Arthur at the moment when we try to upgrade into a target schema. This PR fixes the test that's in place to avoid dealing with tables that (might) depend on each other.

```
$ arthur.py upgrade --into-schema target_schema_for_qa --only-selected some_schema.some_table
...
021-09-15 14:45:13 - CRITICAL - Something bad happened in the ETL: ETLRuntimeError
relations depend on each other while target schema is in effect
  File "/opt/src/arthur-redshift-etl/python/etl/commands.py", line 81, in execute_or_bail
    yield
  File "/opt/src/arthur-redshift-etl/python/etl/commands.py", line 181, in run_arg_as_command
    args.func(args)
  File "/opt/src/arthur-redshift-etl/python/etl/commands.py", line 1170, in callback
    dry_run=args.dry_run,
  File "/opt/src/arthur-redshift-etl/python/etl/load.py", line 1226, in upgrade_data_warehouse
    raise ETLRuntimeError("relations depend on each other while target schema is in effect")
etl.errors.ETLRuntimeError: relations depend on each other while target schema is in effect
```

(Ultimately, the bug is related to not re-leveling or re-numbering relations.)

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -54,6 +54,7 @@ from datetime import datetime, timedelta
 from functools import partial
 from typing import Any, Dict, List, Optional, Sequence, Set
 
+import funcy
 from psycopg2.extensions import connection  # only for type annotation
 
 import etl
@@ -1208,7 +1209,9 @@ def upgrade_data_warehouse(
     exists to include those immediate views in the upgrade.
 
     If a target schema is provided, then the data is loaded into that schema instead of where the
-    relation would normally land.
+    relation would normally land. Note that a check prevents loading tables from different
+    execution levels in order to avoid loading tables that depend on each other. (In that case,
+    the query would have to be rewritten to use the relocated table.)
 
     This is a callback of a command.
     """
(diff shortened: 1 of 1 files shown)


python/etl/load.py
```diff
@@ -1222,8 +1225,14 @@ def upgrade_data_warehouse(
     if not selected_relations:
         return
 
-    if target_schema and any(relation.execution_level > 1 for relation in selected_relations):
-        raise ETLRuntimeError("relations depend on each other while target schema is in effect")
+    involved_execution_levels = frozenset(
+        funcy.distinct(relation.execution_level for relation in selected_relations)
+    )
+    if target_schema and len(involved_execution_levels) != 1:
```
ynaim94, line 1231: The tables specified in the upgrade might still be independent from each and yet have different levels no?

tvogels01: Yes. that's where you come in. 😂 

This PR is meant as a bugfix. The outcome is that we'll be able to use this feature in most cases. But yes, the test is too restrictive. I'd like to fix the current situation and enable the feature.
1. So what's left in documentation or usage info to be consistent here? Should we simply allow only one relation at a time to begin with? What's easy to describe and to code?
2. Follow-up PR: actually test the relations and see whether they do not depend on each other.

ynaim94 (approved):
LGTM 👍

PR #545, opened 2021-10-11
Set max linelength 100

tvogels01:
This PR changes the max linelength for the codebase to 100.

After changing the settings, the second commit is created using:
```shell
pre-commit run black --all-files
```

Final step is to get rid of any `flake8` errors because `black` doesn't break strings.


thomas-vogels:
Closing for now. I'll reopen with my new GitHub account.

PR #548, opened 2021-10-17
Dump all of exc_info into logs (or skip if invalid)

thomas-vogels:
This PR removes logging that fails to consider exception info and instead barfs with:
```
Exception in thread Thread-10:
  File "/usr/lib64/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/load.py", line 875, in poll_worker
    item.mark_failure(relations, exc_info=False)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/load.py", line 206, in mark_failure
    logger.warning("Failed to build relation '%s':", self.identifier, exc_info=exc_info)
  File "/usr/lib64/python3.7/logging/__init__.py", line 1390, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/lib64/python3.7/logging/__init__.py", line 1514, in _log
    self.handle(record)
  File "/usr/lib64/python3.7/logging/__init__.py", line 1524, in handle
    self.callHandlers(record)
  File "/usr/lib64/python3.7/logging/__init__.py", line 1586, in callHandlers
    hdlr.handle(record)
  File "/usr/lib64/python3.7/logging/__init__.py", line 894, in handle
    self.emit(record)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/watchtower/__init__.py", line 225, in emit
    cwl_message = dict(timestamp=int(message.created * 1000), message=self.format(message))
  File "/usr/lib64/python3.7/logging/__init__.py", line 869, in format
    return fmt.format(record)
  File "/tmp/redshift_etl/venv/lib64/python3.7/site-packages/etl/logs/formatter.py", line 70, in format
    values["exception.class"] = record.exc_info[1].__class__.__name__
TypeError: 'bool' object is not subscriptable
```

The original problem is that we set `exc_info` in a logger call outside of an exception handling. That value (`false`) is then sent to the formatter, much to its surprise.

### DIFF
--- a/python/etl/load.py
+++ b/python/etl/load.py
@@ -872,7 +872,8 @@ def create_source_tables_when_ready(
                             "Poller: Recently failed extract found for '%s', marking as failed.",
                             item.identifier,
                         )
-                        item.mark_failure(relations, exc_info=False)
+                        # We are not inside an exception handling so have no exception info.
+                        item.mark_failure(relations, exc_info=None)
                     # We'll create the relation on success and failure (but skip copy on failure)
                     to_load.put(item)
                     # There should be only one event, but definitely don't queue for loading twice.
--- a/python/etl/logs/formatter.py
+++ b/python/etl/logs/formatter.py
@@ -66,11 +66,9 @@ class JsonFormatter(logging.Formatter):
         if hasattr(record, "metrics"):
             values["metrics"] = record.metrics  # type: ignore
         # Always add exception (value) as a field if exception info is present.
-        # if record.exc_info is not None:
-        #     values["exception.class"] = record.exc_info[1].__class__.__name__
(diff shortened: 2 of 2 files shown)


thomas-vogels:
> One comment but makes sense, so this is what caused the poller to die before finishing right? Also the milestone and the arthur version in setup.py are different.

Correct, the poller dies because the exception isn't handled.

The version is set such that we can merge to `master`. And then we can merge back to `next`, for the next release. Milestones are a little less critical here compare our data warehouse repo.

PR #586, opened 2021-11-09
DENG-1120: Add statement_timeout parameter to Arthur Redshift cluster configuration

ynaim94:
## Description

Adding to Arthur the possibility to set [statement_timeout](https://docs.aws.amazon.com/redshift/latest/dg/r_statement_timeout.html)  when running either Arthur load, upgrade or update. The parameter is setup in `resources.RedshiftCluster.statement_timeout` in the Arthur configuration file. The default 0 means to not timeout.

**Note**: This time includes planning, queueing in workload management (WLM), and execution time. There are other ways to monitor queries, see [WLM query monitoring rules](https://docs.aws.amazon.com/redshift/latest/dg/cm-c-wlm-query-monitoring-rules.html).

### User-visible Changes

Users can decide how long they want their query to run before cancelling them, either by setting `resources.RedshiftCluster.statement_timeout` or by passing the value as argument to `--statement-timeout` when running upgrade, updates and load with the Arthur CLI.

### Internal Changes

Arthur reads the configuration file to get the `statement_timeout` value. It will then run `set statement_timeout to {timeout}` everytime an  Arthur load, upgrade or update command is run, which includes all the ETL runs and commands used during development. 

## Links

https://harrys.atlassian.net/browse/DENG-1120

### Values of statement timeout is set correctly in the functions
```
$ arthur.py upgrade --only-selected if__dtc.purchase__tax_lines --statement-timeout 6000
...
2021-11-10 16:20:09 - INFO - Using 10 WLM queue slot(s)
2021-11-10 16:20:09 - INFO - Setting timeout for statements running in Redshift to 6000 ms
...
```

**Note:** 3600000 is the default value for our dev cluster, see: harrystech/harrys-data-warehouse#1858
```
$ arthur.py upgrade --only-selected if__dtc.purchase__tax_lines
2021-11-10 16:21:42 - INFO - Using 10 WLM queue slot(s)
2021-11-10 16:21:42 - INFO - Setting timeout for statements running in Redshift to 3600000 ms
```

### Effect of Setting the timeout on queries
```
$ arthur.py upgrade --only-selected if__dtc.purchase__tax_lines --statement-timeout 600 -q
2021-11-10 16:13:10 - WARNING - Failed upgrade step for 'if__dtc.purchase__tax_lines' (3.55s)                                                                                                                                                 
2021-11-10 16:13:10 - CRITICAL - Something bad happened in the ETL: RequiredRelationLoadError
```

```
$ arthur.py upgrade --only-selected if__dtc.purchase__tax_lines --statement-timeout 6000 -q                                                                                                                    
$
```

### Update

```
(aws:data-dev, prefix:youssef) $ arthur.py update --only-selected if__dtc.purchase__tax_lines --statement-timeout 600
...
2021-11-10 16:34:42 - INFO - Using 10 WLM queue slot(s)
2021-11-10 16:34:42 - INFO - Setting timeout for statements running in Redshift to 600 ms
...
2021-11-10 16:35:17 - WARNING - Failed update step for 'if__dtc.purchase__tax_lines' (0.95s)                                                                                                                                                  
2021-11-10 16:35:17 - CRITICAL - Something terrible happened:
```

```
(aws:data-dev, prefix:youssef) $ arthur.py update --only-selected if__dtc.purchase__tax_lines
...
2021-11-10 16:33:43 - INFO - Setting timeout for statements running in Redshift to 3600000 ms
...
2021-11-10 16:33:56 - INFO - Finished update step for 'if__dtc.purchase__tax_lines' (12.34s)
2021-11-10 16:33:56 - INFO - Ran 'update' for 33.37s and finished successfully!
```

### Load

```
$ arthur.py load -h 
...
 -t MILLISECS, --statement-timeout MILLISECS
                              set the timeout before canceling a statement in Redshift. This time includes planning, queueing in WLM, and execution time. (overrides 'resources.RedshiftCluster.statement_timeout')
...
```

## Deploy Notes

Optional: Setting the value in `aws.yaml` for the data warehouse:

```yaml
 "resources": {
      ...
        "RedshiftCluster": {
            "max_concurrency": 6,
            "wlm_query_slots": 10,
            "statement_timeout": 3600000 # 1 hour, overrides original Redshift cluster parameter value
        }
    }
```

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -464,7 +464,7 @@ def add_standard_arguments(parser: argparse.ArgumentParser, option_names: Iterab
         parser.add_argument(
             "-t",
             "--statement-timeout",
-            metavar="N",
+            metavar="MILLISECS",
             type=int,
             help="set the timeout before canceling a statement in Redshift. This time includes planning,"
             " queueing in WLM, and execution time. (overrides "
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -460,6 +460,16 @@ def add_standard_arguments(parser: argparse.ArgumentParser, option_names: Iterab
             help="set the number of Redshift WLM query slots used for transformations"
             " (overrides 'resources.RedshiftCluster.wlm_query_slots')",
         )
+    if "statement-timeout" in options:
+        parser.add_argument(
+            "-t",
+            "--statement-timeout",
+            metavar="N",
+            type=int,
+            help="set the timeout before canceling a statement in Redshift. This time includes planning,"
+            " queueing in WLM, and execution time. (overrides "
+            "'resources.RedshiftCluster.statement_timeout')",
+        )
     if "skip-copy" in options:
         parser.add_argument(
             "-y",
@@ -1055,7 +1065,16 @@ class LoadDataWarehouseCommand(SubCommand):
 
     def add_arguments(self, parser):
         add_standard_arguments(
-            parser, ["pattern", "prefix", "max-concurrency", "wlm-query-slots", "skip-copy", "dry-run"]
+            parser,
+            [
+                "pattern",
+                "prefix",
+                "max-concurrency",
+                "wlm-query-slots",
+                "statement_timeout",
+                "skip-copy",
+                "dry-run",
+            ],
         )
         parser.add_argument(
             "--concurrent-extract",
@@ -1096,6 +1115,9 @@ class LoadDataWarehouseCommand(SubCommand):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
+        statement_timeout = args.statement_timeout or etl.config.get_config_int(
+            "resources.RedshiftCluster.statement_timeout", 0
+        )
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
@@ -1104,6 +1126,7 @@ class LoadDataWarehouseCommand(SubCommand):
             args.pattern,
             max_concurrency=max_concurrency,
             wlm_query_slots=wlm_query_slots,
+            statement_timeout=statement_timeout,
             concurrent_extract=args.concurrent_extract,
             skip_copy=args.skip_copy,
             skip_loading_sources=args.skip_loading_sources,
@@ -1132,6 +1155,7 @@ class UpgradeDataWarehouseCommand(SubCommand):
                 "prefix",
                 "max-concurrency",
                 "wlm-query-slots",
+                "statement-timeout",
                 "continue-from",
                 "skip-copy",
                 "dry-run",
@@ -1185,6 +1209,9 @@ class UpgradeDataWarehouseCommand(SubCommand):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
+        statement_timeout = args.statement_timeout or etl.config.get_config_int(
+            "resources.RedshiftCluster.statement_timeout", 0
+        )
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
@@ -1193,6 +1220,7 @@ class UpgradeDataWarehouseCommand(SubCommand):
             args.pattern,
             max_concurrency=max_concurrency,
             wlm_query_slots=wlm_query_slots,
+            statement_timeout=statement_timeout,
             only_selected=args.only_selected,
             include_immediate_views=args.include_immediate_views,
             continue_from=args.continue_from,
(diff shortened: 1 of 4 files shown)


python/etl/load.py
```diff
@@ -588,11 +588,11 @@ def verify_constraints(conn: connection, relation: LoadableRelation, dry_run=Fal
     statement_template = """
         SELECT DISTINCT
                {columns}
-          FROM {table}
+          FROM {_table}
```
thomas-vogels, line 591: A leading underscore has a semantic meaning in Python. That's why appending an underscore is usually preferred. So `limit_` instead of `_limit`. 

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.:

* Nice overview:
https://medium.com/python-features/naming-conventions-with-underscores-in-python-791251ac7097

python/etl/commands.py
```diff
@@ -460,6 +460,16 @@ def add_standard_arguments(parser: argparse.ArgumentParser, option_names: Iterab
             help="set the number of Redshift WLM query slots used for transformations"
             " (overrides 'resources.RedshiftCluster.wlm_query_slots')",
         )
+    if "statement-timeout" in options:
+        parser.add_argument(
+            "-t",
+            "--statement-timeout",
+            metavar="N",
```
thomas-vogels, line 467: What is the unit for the timeout? (Seconds? Microseconds? Hours?)

Please add in the text. Or maybe you can use the `metavar` to keep the help text short?

Consider something like this (or **better**):
```diff
diff --git a/python/etl/commands.py b/python/etl/commands.py
index 248dc51..3e4931b 100644
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -464,7 +464,7 @@ def add_standard_arguments(parser: argparse.ArgumentParser, option_names: Iterab
         parser.add_argument(
             "-t",
             "--statement-timeout",
-            metavar="N",
+            metavar="MILLISECS",
             type=int,
             help="set the timeout before canceling a statement in Redshift. This time includes planning,"
             " queueing in WLM, and execution time. (overrides "
```

With this output:
```
  -t MILLISECS, --statement-timeout MILLISECS
                              set the timeout before canceling a statement in Redshift. This time includes planning, queueing in WLM, and execution time. (overrides 'resources.RedshiftCluster.statement_timeout')
```

(Separately, the lowercase `overrides` looks funny after a period. But 🤷 )

python/etl/commands.py
```diff
@@ -1055,7 +1065,16 @@ def __init__(self):
...
     def add_arguments(self, parser):
         add_standard_arguments(
-            parser, ["pattern", "prefix", "max-concurrency", "wlm-query-slots", "skip-copy", "dry-run"]
+            parser,
+            [
+                "pattern",
+                "prefix",
+                "max-concurrency",
+                "wlm-query-slots",
+                "statement_timeout",
```
thomas-vogels, line 1074: Careful, this needs to match the standard arg:
`statement_timeout` -> `statement-timeout`

thomas-vogels (changes requested):
Looks great overall. Small bug in `load` args.

Is there any performance impact? I'm assuming no since this is done per connection and not per relation. But I wanted to check.

python/etl/commands.py
ynaim94: Obviously the only one I didn't test 😅  Thanks, that saved us an ETL run.

thomas-vogels:
Do I see this correctly that this is best for interactive use?

For regular ETLs, I'm assuming that you'd I'd think that using a configuration file (`{"resources": {"RedshiftCluster": {"statement_timeout": ...}}`) is more maintainable. It's called out as "optional" but seems to be easier overall.

ynaim94:
@thomas-vogels You mean in the deploy notes? I meant by that that it is an optional configuration, that Arthur would work even if the value is missing. Let me know if I'm misunderstanding.

ynaim94:
> Is there any performance impact? I'm assuming no since this is done per connection and not per relation. But I wanted to check.

I didn't find anything talking about performance issues in the documentation. This will be ran once every connection is established (once every ETL then?).

thomas-vogels (approved):
LGTM -- Thanks!

thomas-vogels:
> > Is there any performance impact? I'm assuming no since this is done per connection and not per relation. But I wanted to check.
> 
> I didn't find anything talking about performance issues in the documentation. This will be ran once every connection is established (once every ETL then?).

Let's say you have 1000 relations and you add a step that takes 1 s. Now your ETL runs 15 minutes longer. So it's worth keeping in mind how additional changes affect the total ETL time.
Again, shouldn't be a problem here.

More important: Say the cluster has a timeout of 2 hours and I pass in a timeout of 0 using the new command line arg. Does the cluster setting still apply or would the query now run without a timeout?

python/etl/commands.py
```diff
@@ -1096,6 +1127,9 @@ def callback(self, args):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
+        statement_timeout = args.statement_timeout or etl.config.get_config_int(
+            "resources.RedshiftCluster.statement_timeout", 0
+        )
```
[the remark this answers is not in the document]
thomas-vogels: So that feels like a surprise to users. My expectation as a user would be:
* If I specify the statement timeout on the command line, and that's used.
* If I specify the statement timeout in a settings file (and not on the command line), that's used.
* If there's nothing set on the command line or a settings file, the statement timeout from the cluster is used.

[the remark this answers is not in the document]
ynaim94: Sounds good, I will open a PR to use the cluster setup if nothing is set.

PR #601, opened 2021-11-18
DENG-1120: Use default cluster statement timeout value when none is provided

ynaim94:
## Description

Use the default cluster value when no statement timeout is given instead of using zero (no timeout).

## QA

With nothing set in `aws.yaml`:
```
(aws:data-dev, prefix:youssef) $ arthur.py update --only augur_inputs.conversion_v2
...
2021-11-19 18:11:28 - DEBUG - Using default cluster statement timeout
2021-11-19 18:11:28 - INFO - Starting upgrade step for 'augur_inputs.conversion_v2' (1/1)
```

With timeout set in `aws.yaml`:
```
(aws:data-dev, prefix:youssef) $ arthur.py update --only augur_inputs.conversion_v2
...
2021-11-19 18:17:36 - INFO - Using 10 WLM queue slot(s)
2021-11-19 18:17:36 - INFO - Setting timeout for statements running in Redshift to 3600000 ms
...
```

With timeout set in command line:
```
(aws:data-dev, prefix:youssef) $ arthur.py upgrade --only augur_inputs.conversion_v2 -t 1
...
2021-11-19 18:07:21 - INFO - Using 10 WLM queue slot(s)
2021-11-19 18:07:21 - INFO - Setting timeout for statements running in Redshift to 1 ms
2021-11-19 18:07:21 - INFO - Starting upgrade step for 'augur_inputs.conversion_v2' (1/1)
2021-11-19 18:07:21 - WARNING - Failed upgrade step for 'augur_inputs.conversion_v2' (0.06s)...
```

with timeout 0:
```
(aws:data-dev, prefix:youssef) $ arthur.py upgrade --only augur_inputs.conversion_v2 -t 0
...
2021-11-19 18:18:49 - INFO - Using 10 WLM queue slot(s)
2021-11-19 18:18:49 - INFO - Setting timeout for statements running in Redshift to 0 ms
...
```

### DIFF
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -1128,10 +1128,10 @@ class LoadDataWarehouseCommand(SubCommand):
             "resources.RedshiftCluster.max_concurrency", 1
         )
         statement_timeout = args.statement_timeout
-        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
-            statement_timeout = statement_timeout or etl.config.get_config_int(
-                "resources.RedshiftCluster.statement_timeout"
-            )
+        if statement_timeout is None and etl.config.is_config_set(
+            "resources.RedshiftCluster.statement_timeout"
+        ):
+            statement_timeout = etl.config.get_config_int("resources.RedshiftCluster.statement_timeout")
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
@@ -1224,10 +1224,10 @@ class UpgradeDataWarehouseCommand(SubCommand):
             "resources.RedshiftCluster.max_concurrency", 1
         )
         statement_timeout = args.statement_timeout
-        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
-            statement_timeout = statement_timeout or etl.config.get_config_int(
-                "resources.RedshiftCluster.statement_timeout"
-            )
+        if statement_timeout is None and etl.config.is_config_set(
+            "resources.RedshiftCluster.statement_timeout"
+        ):
+            statement_timeout = etl.config.get_config_int("resources.RedshiftCluster.statement_timeout")
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
@@ -1287,10 +1287,10 @@ class UpdateDataWarehouseCommand(SubCommand):
     def callback(self, args):
         relations = self.find_relation_descriptions(args, default_scheme="s3", return_all=True)
         statement_timeout = args.statement_timeout
-        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
-            statement_timeout = statement_timeout or etl.config.get_config_int(
-                "resources.RedshiftCluster.statement_timeout"
-            )
+        if statement_timeout is None and etl.config.is_config_set(
+            "resources.RedshiftCluster.statement_timeout"
+        ):
+            statement_timeout = etl.config.get_config_int("resources.RedshiftCluster.statement_timeout")
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -1127,9 +1127,11 @@ class LoadDataWarehouseCommand(SubCommand):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
-        statement_timeout = args.statement_timeout or etl.config.get_config_int(
-            "resources.RedshiftCluster.statement_timeout", 0
-        )
+        statement_timeout = args.statement_timeout
+        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
+            statement_timeout = statement_timeout or etl.config.get_config_int(
+                "resources.RedshiftCluster.statement_timeout"
+            )
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
@@ -1221,9 +1223,11 @@ class UpgradeDataWarehouseCommand(SubCommand):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
-        statement_timeout = args.statement_timeout or etl.config.get_config_int(
-            "resources.RedshiftCluster.statement_timeout", 0
-        )
+        statement_timeout = args.statement_timeout
+        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
+            statement_timeout = statement_timeout or etl.config.get_config_int(
+                "resources.RedshiftCluster.statement_timeout"
+            )
         wlm_query_slots = args.wlm_query_slots or etl.config.get_config_int(
             "resources.RedshiftCluster.wlm_query_slots", 1
         )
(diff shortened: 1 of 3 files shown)


python/etl/commands.py
```diff
@@ -1127,9 +1127,11 @@ def callback(self, args):
...
             "resources.RedshiftCluster.max_concurrency", 1
         )
-        statement_timeout = args.statement_timeout or etl.config.get_config_int(
-            "resources.RedshiftCluster.statement_timeout", 0
-        )
+        statement_timeout = args.statement_timeout
+        if etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
+            statement_timeout = statement_timeout or etl.config.get_config_int(
+                "resources.RedshiftCluster.statement_timeout"
+            )
```
thomas-vogels, line 1134: When typing this:
```
arthur.py load --statement-timeout 0
```
I would expect the timeout to be set to 0 (effectively, disabled) since I bothered to set something on the commandline. The code however falls back to the setting.

```suggestion
        if statement_timeout is None and etl.config.is_set("resources.RedshiftCluster.statement_timeout"):
            statement_timeout = etl.config.get_config_int(
                "resources.RedshiftCluster.statement_timeout"
            )
```

Also, `get_config_int` errors out with a default of None. This simplifies type checking a lot. If we had an additional `get_config_optional_int` though, the code would simplify further.

python/etl/config/__init__.py
```diff
@@ -338,6 +337,11 @@ def gather_setting_files(config_files: Iterable[str]) -> List[str]:
     return sorted(settings_with_path)
 
 
+def is_set(name: str):
```
thomas-vogels, line 340: This function needs a return type, preferably `bool`.

Overall, the functions here all have `config` in their name, so a `is_config_set` would be more obvious.

Also, all the config-related functions are at the top of the file.

structure

python/etl/config/__init__.py
```diff
@@ -338,6 +337,11 @@ def gather_setting_files(config_files: Iterable[str]) -> List[str]:
...
+    if _mapped_config is not None:
+        return name in _mapped_config.keys()
+
```
thomas-vogels, line 343: What would be the return value if the configuration hasn't been set yet? Or, why wouldn't that be an error?

(FWIW, there some `assert`s in here that I'm not proud of. There should be specific errors raised instead.)

python/etl/dialect/redshift.py
```diff
@@ -535,13 +535,16 @@ def set_wlm_slots(conn: Connection, slots: int, dry_run: bool) -> None:
...
 
-def set_statement_timeout(conn: Connection, statement_timeout: int, dry_run: bool) -> None:
-    etl.db.run(
-        conn,
-        f"Setting timeout for statements running in Redshift to {statement_timeout} ms",
-        f"SET statement_timeout TO {statement_timeout}",
-        dry_run=dry_run,
-    )
+def set_statement_timeout(conn: Connection, statement_timeout: Optional[int], dry_run: bool) -> None:
+    if statement_timeout is None:
```
thomas-vogels, line 539: Style-wise, I've come around to:
```
    if statement_timeout is None:
        logger.info("Using default cluster statement timeout")
        return
```

Which then means that the main code that folks should pay attention to isn't indented (too far).

More importantly, could we use:
```
        logger.debug("Leaving default cluster statement timeout")
```

thomas-vogels (commented):
I'd like to see that this has no timeout:
```
arthur.py load --statement-timeout 0
```
or that it's documented that it's not possible to turn off the timeout.

python/etl/config/__init__.py
ynaim94: Sounds good 👍  How did you generate that view btw?

python/etl/config/__init__.py
ynaim94: It wouldn't be error since the `statement_timeout` is optional in `aws.yaml`, so return `False` if it's not set, which I can add explicitly.

python/etl/dialect/redshift.py
ynaim94: > Style-wise, I've come around to:
>
>   if statement_timeout is None:
>        logger.info("Using default cluster statement timeout")
>        return
> Which then means that the main code that folks should pay attention to isn't indented (too far).

I usually follow this style too but I don't know how confusing for a reader to see a return in a setter.

python/etl/config/__init__.py
thomas-vogels: Once you add `-> bool:` as the return type, there should be an error from `mypy` unless you add an explicit `return False`. 🤞

python/etl/dialect/redshift.py
thomas-vogels: Do you mean because there's no explicit return value? Should be ok for readers.

python/etl/config/__init__.py
thomas-vogels: That's the "Structure" view in PyCharm. (On the left border, below Project.)

python/etl/config/__init__.py
ynaim94: It happened indeed, I added a `return False`

python/etl/dialect/redshift.py
ynaim94: No, because a setter does not return (I'm probably overgeneralizing from my Java 101 😅  ).

ynaim94:
@thomas-vogels Any suggestion on testing these change to commands going forward? It seems like I could have spent more time enumerating edge cases and testing them (Like I'm doing the QA above) but I don't have an idea of how to do that immediately outside an "Arthur + DW" setup environment.

python/etl/dialect/redshift.py
thomas-vogels: Setters are usually related to properties. So I wouldn't call this a setter in Python, but I see where you're coming from with Java.

https://docs.python.org/3/library/functions.html#property

It's good to think about this though ... too many `return`s throw us back into the land of `goto`s :grin: . In small functions the "here's the early exit" will be fine.

thomas-vogels:
@ynaim94 The last example in the PR summary doesn't look right. This should be closer to:

```
arthur.py upgrade --only augur_inputs.conversion_v2 -t 0
```

This could go as-is ... do you want to add any unit tests in a follow-up PR?

ynaim94:
@thomas-vogels Thank you! yes sorry I copy pasted the wrong thing, I tested again with 0. Sounds good for the follow up PR, let me know if you have any suggestions on tests otherwise I can try and come up with something.

thomas-vogels:
Re: testing:  there is a new function, `is_config_set` so you could throw that into a unit test.

And yes, we'd ideally have a demo environment in which we can play through the commands. Goal for 2022?

I don't know whether it's worth it, but we could create a small command ad-hoc in a unit test.

python/etl/commands.py
```diff
@@ -1127,9 +1127,11 @@ def callback(self, args):
         max_concurrency = args.max_concurrency or etl.config.get_config_int(
             "resources.RedshiftCluster.max_concurrency", 1
         )
-        statement_timeout = args.statement_timeout or etl.config.get_config_int(
-            "resources.RedshiftCluster.statement_timeout", 0
-        )
+        statement_timeout = args.statement_timeout
```
thomas-vogels, line 1130: This is basically a "coalesce" operation.

Maybe in a refactor PR, we could have something _like_ this:

```diff
diff --git a/python/etl/config/__init__.py b/python/etl/config/__init__.py
index 11890f0..50ece47 100644
--- a/python/etl/config/__init__.py
+++ b/python/etl/config/__init__.py
@@ -115,6 +115,14 @@ def get_config_list(name: str) -> List[int]:
     return list(map(int, value.split(",")))
 
 
+def coalesce_config_values(latest_value, setting_name, default_value: Optional[str] = None) -> Optional[str]:
+    if latest_value is not None:
+        return latest_value
+    if is_config_set(setting_name):
+        return get_config_value(setting_name)
+    return default_value
+
+
 def set_config_value(name: str, value: str) -> None:
     """Set configuration value to given string."""
     assert _mapped_config is not None, "attempted to set config value before reading config map"
```

Usage is something like `coalesce_config_values(args.statement_timeout, "resources.RedshiftCluster.statement_timeout")` (but adapt to returning `Optional[int]`?).

This gets rid of code duplication and opens up these lines to unit testing. Food for thought.

thomas-vogels (approved):
LGTM

PR #623, opened 2021-12-20
Check for Docker version >= 20

thomas-vogels:
### User-visible Changes

If somebody has an old version of Docker (< 20), then an error will pop up around:
```
docker \
    --pull always
```
We want that feature so that we can check always whether a new version of Arthur has been published on GitHub's packages registry.

Now `bin/arthur.sh` checks the Docker version and advises to upgrade.

## Testing

Testing is a bit tricky unless you really want to downgrade the installed Docker version. But you can easily replace the `docker version` command with an `echo` statement.

### DIFF
--- a/bin/arthur.sh
+++ b/bin/arthur.sh
@@ -14,6 +14,12 @@ if ! type docker >/dev/null ; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


williamknighting (approved):
LGTM :)

ynaim94 (approved):
LGTM

PR #650, opened 2022-02-01
Make Upgrade pipeline on demand

ynaim94:


thomas-vogels:
The "upgrade" pipeline was meant to allow one-off upgrades which might take too long to run locally. In particular the expectation is that the tables would change for this pipeline. What would be the goal of making this an "on-demand" pipeline? Would we set this up for development and then keep re-triggering?

ynaim94:
My bad this was a POC for the on-demand pipelines it wasn't meant to be implemented, I'm closing it now.

PR #657, opened 2022-02-07
DENG-1270: Add On-Demand Rebuild Pipeline

swong28:
### User-visible Changes

Users are able to build a ondemand rebuild pipeline in arthur.

## Links
https://harrys.atlassian.net/browse/DENG-1270

## Testing
Ran below command in arthur
```shell
install_ondemand_rebuild_pipeline.sh jwong
```

Was able to create a on-demand rebuild pipeline:
```
+ aws datapipeline put-pipeline-definition --pipeline-definition file:///tmp/pipeline_definition_arthur_33.json --parameter-values myTimeout=6 --pipeline-id df-03035043C5I4J7VUW9D3
{
    "validationErrors": [], 
    "errored": false, 
    "validationWarnings": []
}
+ set +o xtrace

Your On-Demand Rebuild Pipeline ('df-03035043C5I4J7VUW9D3') has been created!
You can start the pipeline by running the following command:
    aws datapipeline activate-pipeline --pipeline-id df-03035043C5I4J7VUW9D3
```

`df-03035043C5I4J7VUW9D3` is shown in the dev datapipeline:
https://console.aws.amazon.com/datapipeline/home?region=us-east-1#

Was able to kick off on demand rebuild pipeline by running `aws datapipeline activate-pipeline --pipeline-id df-03035043C5I4J7VUW9D3` 

I also got an expected error to make sure we are not concurrently running the pipeline when activating the same pipeline while it is already running:

```
(aws:data-dev, prefix:jwong) $ aws datapipeline activate-pipeline --pipeline-id df-03035043C5I4J7VUW9D3

An error occurred (InvalidRequestException) when calling the ActivatePipeline operation: Web service limit exceeded: Exceeded number of concurrent executions.  Please set the field 'maxActiveInstances' to a higher value in your pipeline or wait for the currenly running executions to complete before trying again
```

### DIFF
--- a/python/etl/templates/text/ondemand_rebuild_pipeline.json
+++ b/python/etl/templates/text/ondemand_rebuild_pipeline.json
@@ -9,7 +9,8 @@
             "role": "${resources.DataPipeline.role}",
             "pipelineLogUri": "s3://${object_store.s3.bucket_name}/_logs/${object_store.s3.prefix}/",
             "region": "${resources.VPC.region}",
-            "maximumRetries": "2"
+            "maximumRetries": "2",
+            "maxActiveInstances": "1"
         },
         {
             "id": "SNSParent",
--- a/python/etl/templates/text/ondemand_rebuild_pipeline.json
+++ b/python/etl/templates/text/ondemand_rebuild_pipeline.json
@@ -20,21 +20,21 @@
             "id": "SuccessNotification",
             "type": "SnsAlarm",
             "parent": { "ref": "SNSParent" },
-            "subject": "ETL Rebuild Success: ${object_store.s3.prefix}",
+            "subject": "ETL On-Demand Rebuild Success: ${object_store.s3.prefix}",
             "message": "Completed last action successfully at #{node.@actualEndTime}\\nLast node: #{node.name}\\nPipelineId: #{node.@pipelineId}\\nLog directory: #{node.pipelineLogUri}#{node.@pipelineId}"
         },
         {
             "id": "FailureNotification",
             "type": "SnsAlarm",
             "parent": { "ref": "SNSParent" },
-            "subject": "ETL Rebuild Failure: ${object_store.s3.prefix}",
+            "subject": "ETL On-Demand Rebuild Failure: ${object_store.s3.prefix}",
             "message": "Failed step #{node.name} at #{node.@actualEndTime}\\nPipelineId: #{node.@pipelineId}\\nLog directory: #{node.pipelineLogUri}#{node.@pipelineId}\\n\\nShow error message: arthur.py show_pipelines '#{node.@pipelineId}'"
         },
         {
             "id": "PagerNotification",
             "type": "SnsAlarm",
             "topicArn": "arn:aws:sns:${resources.VPC.region}:${resources.VPC.account}:${resource_prefix}-page",
-            "subject": "ETL Rebuild Failure: ${object_store.s3.prefix}",
+            "subject": "ETL On-Demand Rebuild Failure: ${object_store.s3.prefix}",
             "message": "Failed step #{node.name} at #{node.@actualEndTime}\\nPipelineId: #{node.@pipelineId}\\nLog directory: #{node.pipelineLogUri}#{node.@pipelineId}\\n\\nShow error message: arthur.py show_pipelines '#{node.@pipelineId}'"
         },
         {
--- a/python/etl/templates/text/ondemand_rebuild_pipeline.json
+++ b/python/etl/templates/text/ondemand_rebuild_pipeline.json
@@ -1,0 +1,184 @@
+{
+    "objects": [
+        {
+            "id": "Default",
+            "name": "Default",
+            "scheduleType": "ondemand",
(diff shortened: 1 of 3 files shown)


python/etl/templates/text/ondemand_rebuild_pipeline.json
```diff
@@ -0,0 +1,184 @@
...
+        {
+            "id": "Default",
+            "name": "Default",
+            "scheduleType": "ondemand",
+            "failureAndRerunMode": "CASCADE",
+            "resourceRole": "${resources.EC2.iam_instance_profile}",
+            "role": "${resources.DataPipeline.role}",
+            "pipelineLogUri": "s3://${object_store.s3.bucket_name}/_logs/${object_store.s3.prefix}/",
+            "region": "${resources.VPC.region}",
+            "maximumRetries": "2"
```
ynaim94, line 12: Should we add `maxActiveInstances` to be 1 in case Airflow runs a task twice for any reason?

ynaim94 (commented):
One comment otherwise LGTM

ynaim94:
@swong28 I'm getting this: 
```
$ aws datapipeline activate-pipeline --pipeline-id df-03035043C5I4J7VUW9D3

An error occurred (InvalidRequestException) when calling the ActivatePipeline operation: Pipeline definition has errors: Cannot activate a pipeline with errors
```

I wonder if it is because shortly I tried to activate it from the UI? I'll try deploying a new one.

ynaim94:
@swong28 Another question: I was expecting the command to create a new pipeline each time I run it, but it just keeps overwriting the same one. Is that the expected behavior?

ynaim94:
@swong28 I guess I don't understand why is it giving the same pipeline ID, whereas something as an `install_upgrade_pipeline` and `install_validation_pipeline` will create a new one everytime. Where is the difference in the code?

swong28:
> @swong28 I guess I don't understand why is it giving the same pipeline ID, whereas something as an `install_upgrade_pipeline` and `install_validation_pipeline` will create a new one everytime. Where is the difference in the code?

It is because the start date is included in the pipeline creation clis as an argument. 

like this:
```
aws datapipeline create-pipeline \
    --unique-id dw-etl-upgrade-pipeline \
    --name "$PIPELINE_NAME" \
    --tags $AWS_TAGS \
    | tee "$PIPELINE_ID_FILE"
```

And see below documentation for `aws datapipeline create-pipeline`
```
For example, if the first call to CreatePipeline does not succeed, you can pass in the same unique identifier and pipeline name combination on a subsequent call to CreatePipeline . CreatePipeline ensures that if a pipeline already exists with the same name and unique identifier, a new pipeline is not created.
```
https://docs.aws.amazon.com/cli/latest/reference/datapipeline/create-pipeline.html

ynaim94:
@swong28 Sounds good thanks for checking, so what happens when we run [put_pipeline_definition](https://docs.aws.amazon.com/datapipeline/latest/APIReference/API_PutPipelineDefinition.html) to an existing populated pipeline? From the documentation, it sounds like it `appends` instead of overriding the existing one.

If we want to get away from this, should we just recreate a new pipeline every time we run `install_ondemand_rebuild_pipeline`? Recreating from scratch assures nothing funky is going on, and I believe it's also how these scripts `install_...` are intended to be used.

swong28:
> @swong28 Sounds good thanks for checking, so what happens when we run [put_pipeline_definition](https://docs.aws.amazon.com/datapipeline/latest/APIReference/API_PutPipelineDefinition.html) to an existing populated pipeline? From the documentation, it sounds like it `appends` instead of overriding the existing one, it's not super clear.
> 
> If we want to get away from this, should we just recreate a new pipeline every time we run `install_ondemand_rebuild_pipeline`? Recreating from scratch assures nothing funky is going on, and I believe it's also how these scripts `install_...` are intended to be used.
> 
> Edit: I also don't really see a value in idempotency here in term of pipeline id returned (it could be useful for other cases), I care more about correctness of the pipeline and it following the definition we laid out in the Json everytime.

That sounds good too! I will modify the name to have a timestamp to ensure a new pipeline will be created each time.

python/etl/templates/text/ondemand_rebuild_pipeline.json
```diff
@@ -0,0 +1,185 @@
...
+        },
+        {
+            "id": "SNSParent",
+            "topicArn": "arn:aws:sns:${resources.VPC.region}:${resources.VPC.account}:${resource_prefix}-status"
+        },
+        {
+            "id": "SuccessNotification",
+            "type": "SnsAlarm",
+            "parent": { "ref": "SNSParent" },
+            "subject": "ETL Rebuild Success: ${object_store.s3.prefix}",
```
ynaim94, line 23: Do you think it would make more sense to change the name to ETL on-demand?

ynaim94 (approved):
LGTM, but one comment around naming the pipeline

python/etl/templates/text/ondemand_rebuild_pipeline.json
swong28: I should change the name!

PR #674, opened 2022-03-09
Remove unload from on-demand rebuild pipeline

ynaim94:
## Description

This PR removes the unload from the ondemand pipeline. We don't want it there since we want to use this pipeline to run the  Arthur load and transform in staging only, so we only want to unload after the promote, which does not happen in this pipeline.

~This is very specific to our use case of the migration so one way to make this backward compatible (since this is open-source) is to actually create another pipeline that would be called "staging_rebuild" or something like that.~ 
See Tom's comment: https://github.com/harrystech/arthur-redshift-etl/pull/674#issuecomment-1063246432

## Test

We have a full rebuild run yesterday in dev: https://us-east-1.console.aws.amazon.com/datapipeline/home?region=us-east-1#ExecutionDetailsPlace:pipelineId=df-06381941BFK4A6JTL0WQ&show=latest

### DIFF
--- a/python/etl/templates/text/ondemand_rebuild_pipeline.json
+++ b/python/etl/templates/text/ondemand_rebuild_pipeline.json
@@ -145,22 +145,13 @@
             "dependsOn": { "ref": "ArthurLoad" }
         },
         {
-            "id": "ArthurUnload",
-            "name": "Arthur Unload (EC2)",
-            "type": "ShellCommandActivity",
-            "parent": {"ref": "ArthurCommandParent"},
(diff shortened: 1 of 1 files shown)


thomas-vogels:
I'd think of the rebuild and refresh pipelines as the official interface. So I'm not worried about removing unload from the "on demand" pipeline. 
(We haven't described how to leverage the "on demand" pipeline in this repo, have we? Without that update to our docs, I'd still consider the "on demand" pipeline as not GA.)

swong28 (approved):
LGTM

PR #698, opened 2022-04-07
Set relation_column_encoding to AUTO by default

### LINKED ISSUE #229: Add default encodings
# Summary

In most cases, we should be able to pick a reasonable default for an encoding. Setting an encoding prevents us from having to have Redshift cluster figure it out (`COMPUPDATE  ON`) or override our choices (`COMPUPDATE PRESET`) which saves time during load and reduces storage.

# Details

* Use `raw` for columns in distribution, sort, and foreign keys
* Use `raw` for boolean, real, double-precision
* Use `bytedict` for statuses and and anything enum-based upstream
* Use `az64` for int, date, timestamps
* Else use `zstd`

thomas-vogels:
### User-visible Changes

This changes the default of the `relation_column_encoding` setting from `ON` to `AUTO`. This means that Arthur will pick appropriate encodings for columns. (The encoding value is picked based on the type and whether it's used in distribution or sort keys.) Selecting the encoding speeds up loading data (`COMPUPDATE` can be left off).

## Links

Closes #229

## Testing

You can check the active setting using:
```
arthur.py settings arthur_settings.redshift.relation_column_encoding
```

You can check which encoding is used with `show_ddl` on a table.  Unless you already had the auto setting active, you would see for example columns like this:
```
    "created_at" TIMESTAMP WITHOUT TIME ZONE NOT NULL,
```
but then would see an encoding with the new default:
```
    "created_at" TIMESTAMP WITHOUT TIME ZONE ENCODE zstd NOT NULL,
```

## Deploy Notes

If the `relation_column_encoding` is set to `AUTO` in local configurations, then this can be removed now. If this mode is not desired, then `relation_column_encoding` must be set to `OFF` now.

Harry's internal note: This has been effectively in place in production given that we have a file in the object store with this content:
```yaml
{
  "arthur_settings": {
    "redshift": {
       "relation_column_encoding": "AUTO"
    }
  }
}
```
(This file should now be removed.)

### DIFF
--- a/python/etl/config/default_settings.yaml
+++ b/python/etl/config/default_settings.yaml
@@ -16,7 +16,7 @@
     "retriable_error_codes": "8001,15001,15005",
     "concurrent_load_idle_termination_seconds": 3600,
     "redshift": {
-      "relation_column_encoding": "ON"
+      "relation_column_encoding": "AUTO"
     }
   },
   # Target (Redshift) cluster
--- a/python/etl/dialect/redshift.py
+++ b/python/etl/dialect/redshift.py
@@ -181,9 +181,7 @@ def add_auto_encoding(table_design: dict) -> None:
 
 def build_table_ddl(table_name: TableName, table_design: dict, is_temp=False) -> str:
     """Assemble the DDL of a table in a Redshift data warehouse."""
-    if (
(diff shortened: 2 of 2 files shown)


ynaim94 (approved):
LGTM

PR #728, opened 2022-05-27
DENG-1806: Show downstream dependents in dbt

ynaim94:
## Description

This changes allow us to use the DBT image in `$dw/dbt/` to run DBT commands directly from Arthur commands. The implmentation in this story is for `show_downstream_dependents`. We can now add a flag `--include-dbt` to show downstream dependants in `dbt` of the tables specified in the command.

The changes are:

* Arthur assumes there’s a folder `/dbt` in your data warehouse path with a dockerfile and a dbt project if you use the `--with-dbt` flag. The docker image can take commands, run dbt commands and return the `stdout` to Arthur.

So this is not breaking the existing interface per se because Arthur would still work the same if you don’t need/have dbt. 

### Harry's only usecase

Users can run the normal show_downstream_dependents:

```
(aws:data-dev, prefix:youssef) # arthur.py show_downstream_dependents augur_outputs -q
augur_outputs.results      # DATABASE index=   1 level=  1 flag=selected  is_required=False                                                                                                                                                   
augur_outputs.results_next # DATABASE index=   2 level=  1 flag=selected  is_required=False
```

Or they can use the `--include-dbt` flag (notice the last line is a dbt table):

```
(aws:data-dev, prefix:development) $ arthur.py show_downstream_dependents augur_outputs -q --include-dbt
augur_outputs.results      # DATABASE index=   1 level=  1 flag=selected  is_required=False                                                                                                                                                   
augur_outputs.results_next # DATABASE index=   2 level=  1 flag=selected  is_required=False
ods-full_funnel_metrics # TABLE index=   1 flag=DBT       is_required=True
```

## Links

https://harrys.atlassian.net/browse/DENG-1806

## Testing

Adding `--include-dbt` to `(aws:data-dev, prefix:youssef) # arthur.py show_downstream_dependents augur_outputs -q --with-dbt`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,25 +20,24 @@ RUN yum install -y \
         jq \
         libyaml-devel \
         openssh-clients \
+        passwd \
         postgresql \
         procps-ng \
         python3 \
         python3-devel \
+        sudo \
         tar \
         tmux \
-        vim-minimal \
-        sudo \
-        passwd
+        vim-minimal
 
 RUN amazon-linux-extras install docker
 
 # Run as non-priviledged user "arthur".
-RUN useradd --comment 'Arthur ETL' --user-group --create-home arthur && \
+RUN useradd --comment 'Arthur ETL' --user-group --groups wheel --create-home arthur  && \
     mkdir --parent /opt/data-warehouse "/opt/local/$PROJ_NAME" /opt/src/arthur-redshift-etl && \
     chown -R arthur.arthur /opt/*
 
 RUN echo "arthur:arthur" | chpasswd
-RUN usermod -aG wheel arthur
 
 USER arthur
 
--- a/python/etl/commands.py
+++ b/python/etl/commands.py
@@ -1792,7 +1792,11 @@ class ShowDownstreamDependentsCommand(SubCommand):
         dbt_downstream_parents = " ".join(
             [f"{parent}+" for parent in dbt_project.show_downstream_dbt_parents(dbt_model_identifiers)]
         )
+        if not dbt_downstream_parents:
+            logging.info("No dbt downstream dependents found")
+            return
 
+        logging.info("dbt downstream dependents found")
         dbt_project.build_image()
         dbt_stdout = dbt_project.run_cmd(
             f"dbt list -t dev --exclude redshift --output json "


Dockerfile
```diff
@@ -28,18 +28,17 @@ RUN yum install -y \
...
         vim-minimal
 
-# Run as non-priviledged user "arthur".
-RUN useradd --comment 'Arthur ETL' --user-group --create-home arthur && \
-    mkdir --parent /opt/data-warehouse "/opt/local/$PROJ_NAME" /opt/src/arthur-redshift-etl && \
-    chown -R arthur.arthur /opt/*
-USER arthur
+RUN amazon-linux-extras install docker
+
+#  TODO(youssef): We should use a non root user, but the containers need access the docker socket which is owned by root
```
thomas-vogels, line 33: This is a substantial change. Could we make this more temporary and leave the user `arthur` just run as `root`?

thomas-vogels (commented):
@soumyadsanyal I'd normally block this PR.

If we need to run something as `root`, then we should run that thing as `root`. I'd like to know why we have to change the entire setup which now goes against best practices.

Dockerfile
ynaim94: I would prefer this solution too, but I struggled a bit to give root privileges to Arthur in a way I could access the `docker.sock`. As for why I'm going against best practices, I could not find a way to go around this as per description:
```
Arthur runs in a container as root and not as non-privileged user arthur (that’s the only way I could find to have access to the host docker socket which is owned by root on macs).
```

thomas-vogels: The problem that I see is that the socket doesn't get mounted until runtime. So would changing permissions on `/var/run/docker.sock` in the entrypoint script solve this in a way that we can run as the `arthur` user which then can read/write to the socket?

ynaim94: Wouldn't that mean that we actually change the permission of the file in the host machine? And is that possible to do in the entrypoint without sudo? I've tried a bit to go the sudo route and stopped at the container refusing sudo commands because the arthur user is not a `sudoer`. If that's a viable option I could try again.

thomas-vogels: The permissions on the host machine will stay the same. You can try it out by using `USER root` at the end of the `local` section in the Dockerfile. 
The fastest way to make `arthur` a sudoer is to add the user to the `wheel` group. But that might still require a one-time password so we'd have to throw in a `NOPASSWD` for the permission change. https://www.sudo.ws/docs/man/1.8.15/sudoers.man/#NOPASSWD

ynaim94: Makes more sense, I was a bit confused by how docker handles volumes, I guess it points to a copy of the volume and not the volume itself? I also went ahead with modifying the permission in the entrypoint and just creating the arthur user with a password that I reuse in the entrypoint. Let me know if that's better 😅  Thank you

ynaim94:
@thomas-vogels fyi I also made a change to the Sphinx version, seems like it was breaking the docker build in the latest `next`.

requirements-docs.txt
```diff
@@ -1,4 +1,4 @@
 myst-parser~=0.16
-Sphinx~=4.5
+Sphinx==4.3.2
```
thomas-vogels, line 2: Sorry about that.

thomas-vogels (approved):
LGTM -- thanks for the updates.

Good for now -- I'll probably make a suggestion later around passwords. Using `arthur` as the user, group and password might get confusing :-)

PR #850, opened 2023-07-12
add deletion policy parameter

rishi-gajula-harrys:

### DIFF
--- a/cloudformation/dw_cluster.yaml
+++ b/cloudformation/dw_cluster.yaml
@@ -83,6 +83,12 @@ Parameters:
         Description: (optional) Preferred maintenance window for the Redshift cluster
         Type: String
         Default: "sun:18:30-sun:19:00"
+    DeletionPolicyParameter:
+        Description: Deletion policy defaulted to retain so that resources are retained when stack is deleted
+        Type: String
+        AllowedValues:
+        - Retain
+        Default: Retain
 
 Conditions:
 
@@ -117,6 +123,7 @@ Resources:
         Type: "AWS::Redshift::ClusterSubnetGroup"
         Properties:
             Description: "Cross-reference public subnet ID from VPC stack"
+            DeletionPolicy: !Ref DeletionPolicyParameter
             SubnetIds:
                 - Fn::ImportValue: !Sub "${VpcStackName}::public-subnet-id"
             Tags:
@@ -127,6 +134,7 @@ Resources:
         Type: "AWS::Redshift::ClusterParameterGroup"
         Properties:
             Description: "Parameter group for Redshift cluster"
+            DeletionPolicy: !Ref DeletionPolicyParameter
             ParameterGroupFamily: "redshift-1.0"
             Parameters:
                 - ParemeterName: "auto_analyze"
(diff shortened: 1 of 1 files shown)


cloudformation/dw_cluster.yaml
```diff
@@ -115,6 +115,7 @@ Resources:
 
     RedshiftClusterSubnetGroup:
         Type: "AWS::Redshift::ClusterSubnetGroup"
+        DeletionPolicy: Retain
```
jwisdom-harrys, line 118: Can we change this to a `Parameter` with the default value of `Delete` and only the appropriate allowed values? That way, we can specify the parameter to switch to `Retain` for the stack we care about, but the mainline default behavior of this cloudformation stack won't change.

rishi-gajula-harrys: upon checking error looks like it does not like ref function in deletion policy https://stackoverflow.com/questions/34286395/cloudformation-template-format-error-every-deletionpolicy-member-must-be-a-stri

cloudformation/dw_cluster.yaml
```diff
@@ -200,21 +224,34 @@ Resources:
             PreferredMaintenanceWindow:
                 !Ref PreferredMaintenanceWindow
             VpcSecurityGroupIds:
-                - Fn::ImportValue: !Sub "${VpcStackName}::redshift-public-sg"
+                - !Ref AdditionalSecurityGroupParam1
```
jwisdom-harrys, line 227: Nit: I think importing the one from the stack is still the right play, but just adding one additional security group param

rishi-gajula-harrys: yeah right, let me update this.

jwisdom-harrys (approved):

cloudformation/dw_cluster.yaml
```diff
@@ -201,20 +222,33 @@ Resources:
...
+                - !Ref AdditionalSecurityGroupParam
             SnapshotIdentifier:
                 !If [ "HasSnapshotIdentifier", !Ref "SnapshotIdentifier", !Ref "AWS::NoValue" ]
             Tags:
                 - Key: user:sub-project
                   Value: redshift-cluster
+                - Key: user:stack-env-name
+                  Value": dev
+                - Key: hs-environment
+                  Value: dev
```
jwisdom-harrys, line 234: Oops missed these, going to have to add a parameter to specify those

cloudformation/dw_cluster.yaml
```diff
@@ -125,16 +145,17 @@ Resources:
...
-                - ParemeterName: "auto_mv"
-                  ParemeterValue: "true"
-                - ParemeterName: "datestyle	"
-                  ParemeterValue: "ISO, MDY"
+                - ParameterName: "auto_analyze"
+                  ParameterValue: "true"
+                - ParameterName: "auto_mv"
+                  ParameterValue: "true"
+                - ParameterName: "datestyle"
+                  ParameterValue: "ISO, MDY"
```
ynaim94-harrys, line 158: These are values that I'm assuming were not being applied, what changes now that they are?

jwisdom-harrys (approved):

PR #852, opened 2023-07-17
add default for additional security group param

rishi-gajula-harrys:
## Description

added default to empty since prod cluster has only one security group whereas dev has 2 security groups (note: we added this parameter in cloudformation since dev had 2 security groups attached to the cluster)

### DIFF
--- a/cloudformation/dw_cluster.yaml
+++ b/cloudformation/dw_cluster.yaml
@@ -99,6 +99,7 @@ Parameters:
     AdditionalSecurityGroupParam:
         Description: Add security group
         Type: String
+        Default: ""
     Environment:
         Description: specify env dev or prod
         Type: String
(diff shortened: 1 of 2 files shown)


cloudformation/dw_cluster.yaml
```diff
@@ -225,8 +222,7 @@ Resources:
             PreferredMaintenanceWindow:
                 !Ref PreferredMaintenanceWindow
             VpcSecurityGroupIds:
-                - Fn::ImportValue: !Sub "${VpcStackName}::redshift-public-sg"
-                - !Ref AdditionalSecurityGroupParam
+                - sg-eff0db9c
```
thomas-vogels, line 225: This is a public repo. I wouldn't hard-code security groups.

PR #859, opened 2023-08-28
Drop assignee (personal account), fix reviewers

thomas-vogels:
## Description

This drops my personal account as an assignee for PRs from dependabot.
This also fixes the team name to data-platform.

And since `pre-commit` would fail otherwise, there's a version upgrade for `shellcheck`.

### User-visible Changes

None.

### Internal Changes

None.
(This affects the repo, not Arthur).

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,23 +1,19 @@
 version: 2
 
 updates:
-  - assignees:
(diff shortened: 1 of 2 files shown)


ynaim94-harrys (approved):
LGTM

